PR #1, opened 2019-07-31
Implement stub hintr input validation

r-ash:
This PR will
* Implement stub of hintr input validation code
* Only return country from PJNZ file at the moment and always validate as true


R/endpoints.R
```diff
@@ -0,0 +1,46 @@
+api <- function() {
+  pr <- plumber$new()
+  pr$handle("POST", "/validate", handle_validate)
+  pr
+}
+
+handle_validate <- function(req, pjnz, shape, population, survey, programme,
```
richfitz: The naming scheme through here is going to need some work - it might be clearer if you implemented *two* mini endpoints (even if one is just really trivial like the package version).  I think that we need something like:

```
hintr_response <- function(success, value, error) {
}
```

that all your responses will go through (`prepare_success_response` and `prepare_fail_response` *sound* general but are in fact quite specific).  The `handle_validate` function will need a somewhat better name too - perhaps `pjnz_validate`?  But it might also be worth prefixing these with something, so we might have `endpoint_pjnz_validate` and `endpoint_hint_version` for example - then we'd easily see that these are functions that will definitely return things that are acceptable to the serialiser

R/endpoints.R
```diff
@@ -0,0 +1,46 @@
...
+      "anc" = list()
+    )
+  )
+}
+
+prepare_fail_response <- function() {
+  list(
+    "status" = "failure",
+    "errors" = list(
+      "pjnz" = c("EXAMPLE_FAILURE")
```
richfitz: For errors, one thing that might be nice is if we have an enum of error types and return a simple list like `list(type = "InvalidInput", message = "The file was not the expected format")`.  We did that in montagu and when I was consuming the API it made it easier to tool around

DESCRIPTION
```diff
@@ -12,6 +12,10 @@ Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 6.1.1
+Imports:
+    specio
+Remotes:
```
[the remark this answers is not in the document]
r-ash: Thanks, removed that now.

R/endpoints.R
r-ash: I was kind of envisioning a similar thing, something like

```
"errors" : {
  "pjnz": [
    "MISMATCH_COUNTRY_ERROR"
  ],
  "shape": [
    "MISMATCH_COUNTRY_ERROR",
    "MISMATCH_DISTRICTS_ERROR"
  ],
  "population": [
    "MISMATCH_DISTRICTS_ERROR",
  ],
...
```

Where the errors returned are some key which the front end can choose user readable wording for. Is there an advantage to returning a message along with that?

tests/testthat/test_endpoints.R
```diff
@@ -0,0 +1,46 @@
+context("endpoints")
+
+test_that("endpoint_validate_baseline correctly validates data", {
+  pjnz <- system.file("testdata", "Botswana2018.PJNZ", package = "hintr")
+  response <- endpoint_validate_baseline("request", pjnz, NULL, NULL)
+  expected_response <- '{"status":"success","errors":{},"data":"Botswana"}'
+  expect_equal(as.vector(response), expected_response)
```
r-ash: Do you know if there is a better way for me to test the returned JSON? Casting to vector and comparing with a string seems a bit rubbish.

PR #3, opened 2019-08-13
Validate requests and responses using jsonvalidate

r-ash:
This PR will
* Use jsonvalidate to validate incoming requests and outgoing responses against a schema
* Update schema for input validation


tests/testthat/test-endpoints.R
```diff
@@ -1,24 +1,39 @@
 context("endpoints")
 
-test_that("endpoint_validate_baseline correctly validates data", {
+validate_test_that("endpoint_validate_input correctly validates data", {
```
r-ash: I had a question about this. So we're only doing schema validation when an environmental variable is set. What is the best way to make this work? Is there a way to set the env var for all tests from one place?

r-ash:
This PR doesn't include `jsonvalidate` in the `DESCRIPTION` or `.travis.yml`. This is because I was expecting we want to merge that before merging this so I will include it in the build once we've merged. 

Where should that go though? Using `Imports` section in the description pulls from CRAN right? And the updated version takes time to get onto CRAN after it is merged or is that a quick process?

I will take a look at the jsonvalidate PR now but it seems to be working fine. The only issue is it is returning lots of warnings about schema id being ignored

```
test-endpoints.R:12: warning: endpoint_validate_input correctly validates data
schema id ignored

test-endpoints.R:12: warning: endpoint_validate_input correctly validates data
ValidateInputResponse

test-endpoints.R:12: warning: endpoint_validate_input correctly validates data
schema id ignored

test-endpoints.R:12: warning: endpoint_validate_input correctly validates data
ValidateInputResponse
```

Should we just remove the schema id? Or is the schema id required by the validation on the JS side?

tests/testthat/test-endpoints.R
richfitz: I would set the environment variable machine-by-machine: so in your .Renviron, and in the .travis.yml

R/json.R
```diff
@@ -1,14 +1,29 @@
...
-    valid <- TRUE
+    valid <- validate(json, schema)
   } else {
     valid <- TRUE
   }
   valid
 }
 
+validate <- function(json, schema) {
+  schema <- system.file("schema", paste0(schema, ".schema.json"),
```
richfitz: add a `mustWork = TRUE` here (probably worth adding a `system_file <- function(...) system.file(..., mustWork = TRUE)` type wrapper to your utilities)

R/json.R
```diff
@@ -1,14 +1,29 @@
...
+#' is set to true. If VALIDATE_JSON_SCHEMAS is not true then this always
+#' returns TRUE.
+#'
+#' @param json The JSON to validate.
+#' @param schema Name of the schema to validate against.
+#'
+#' @return True if JSON adheres to schema.
+#' @keywords internal
+validate_json_schema <- function(json, schema) {
   if (validate_schemas()) {
```
richfitz: What do you think about rewriting in the happy-path style:

```
if (!validate_schemas()) {
  return(TRUE)
}
validate(json, schema)
```

inst/schema/ValidateInputResponse.schema.json
```diff
@@ -1,11 +1,13 @@
 {
-	"id": "File",
+	"id": "ValidateInputResponse",
```
richfitz: It's up to you, but you can leave these `id`s off as they will be added for you as the schema is built

tests/testthat/helper-endpoint.R
```diff
@@ -20,3 +20,9 @@ build_validate_request <- function(pjnz) {
     )
   )
 }
+
+validate_test_that <- function(desc, code) {
```
richfitz: drop this, see main comment

PR #4, opened 2019-08-13
Update example payload

r-ash:
I missed including this in previous PR.

This updated the example payload so that the instructions in the README for running and testing the docker container actually work.


r-ash:
Build is passing now, I cleared the build cache and I think that solved the problem.

PR #5, opened 2019-08-19
Set up model runner and redis and add integration test

r-ash:
This PR will

* Add code to create a queue
* Add an integration test which creates a queue with runners and runs a job on the queue

Expecting next PR to be
* Expose model run endpoint which will push a model run job onto the queue and return job id
* Expose model status endpoint which will take a job id and return the data if it has completed or return indication of in progress if it has not completed.


docker/build_test_containers
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+
+# Initialise docker containers required for running integration tests
+docker run --rm -d --network=host --name hintr_redis redis
```
r-ash: This is something I wanted to ask about. I have a test `test_integration.R` which needs the redis container to be running to test adding items to the queue. What is the best way to do that? Is it okay to run up a docker container in a script like this?

richfitz: This will be easiest to talk about when you're back, but you can use [`stevedore`](https://github.com/richfitz/stevedore) to control docker from your tests, or we can set the tests up to assume the existence of redis and skip gracefully if it's not available

PR #6, opened 2019-08-20
Fix json response issues

r-ash:
This PR will

* Ensure returned json is not being parsed a string by plumber. This fixes bug where we were seeing all responses as a json array containing a single string
* Return response code 400 should the status be a failure for the validation endpoint

There is a lot of noise in this PR as we're comparing to master. We should merge mrc-400 into master first then this will look more sensible.


example/payload.json
```diff
@@ -1,4 +1,4 @@
 {
 	"type": "pjnz",
-	"path": "src/inst/testdata/Botswana2018.PJNZ"
```
r-ash: Also wondering if there is a better way to work around this in future. Perhaps that will come out of the work setting up docker volume though? Instead of having dupe example payloads, one for when the data is living in docker in a `src` directory and one for when running the plumber API outside of docker.

r-ash:
I've also fixed the coverage problems in this PR.

I think there is definitely still opportunity get improve the unit test coverage and documentation. If anything jumps out as lacking let me know but I'm keen to try and get this merged to unblock Alex and Emma.

example/payload.json
hillalex: This is where a configurable root directory would come in handy!

richfitz:
To fix the testing and expose the problem, change validate to

```r
validate <- function(json, schema) {
  schema <- system_file("schema", paste0(schema, ".schema.json"),
                        package = "hintr")
  jsonvalidate::json_validate(json, schema, engine = "ajv", error = TRUE)
}
```

that is, add `error = TRUE` (or throw your own errors).  By default `jsonvalidate::json_validate` returns `TRUE` / `FALSE`, so this is silently passing

r-ash:
Okay I think this should be good to go now. The validation of the actual data and error types being returned is pretty loose. I figured it would suffice to add some proper validation there once we have actual data to work with?

tests/testthat/test-queue.R
```diff
@@ -0,0 +1,67 @@
...
+  model_queue_finalize(queue)
+  ## Pase for workers to be stopped
+  Sys.sleep(1)
+  expect_equal(queue$worker_len(), 0)
+  expect_equal(queue$worker_log_tail(worker_1)$command, "STOP")
+  expect_equal(queue$worker_log_tail(worker_2)$command, "STOP")
+  expect_equal(global_queue$queue, queue)
+
+  ## queue can be torn down
+  model_queue_stop()
```
r-ash: Another thing I wanted to ask about was this test. 

It returns an error which doesn't fail the build.
```
Stopping workers
Error in con$SMEMBERS(keys$worker_name) : attempt to apply non-function
```

I think this must either be coming from the call to `model_queue_stop()` or `model_queue_finalize()` call, but I'm not sure which or how to avoid it. @richfitz do you know how I can avoid that?

richfitz: let's deal with this this as part of the tidy-up of the model runner code

PR #8, opened 2019-08-21
Add endpoint for running model and querying status

r-ash:
This PR will
* Add an endpoint for running the model
* Add endpoint for querying model run status
* Update schemas

There is lots of noise here again but some of that should be tidied up after we merge mrc-418.


r-ash:
Have reworked this in #22

PR #9, opened 2019-08-22
Write shapefile upload validation + response

r-ash:
This PR will
* Validate shape file can be read and is for 1 country only
* Will store list of regions from the shapefile in redis for later use in validation
* Return geojson to front end 

We should merge mrc-418 into master before this and then this PR will be much tidier.


R/store.R
```diff
@@ -0,0 +1,43 @@
...
+  invisible(global_store$store)
+}
+
+store_set <- function(key, data) {
+  global_store$store$set(key, data)
+}
+
+refresh_store <- function() {
+  if(!is.null(global_store$store) && length(global_store$store$list()) > 0) {
+    global_store$store$clear()
```
r-ash: Looks like tests are still unstable. I'm trying to create a way here to make sure I have a fresh store for each test to use. But looks like this keeps running into problems with restarting the store or properly clearing it. Is there a good way to do this? I've tried using `store$destroy()` too but no luck there either.

r-ash:
In fact this isn't quite ready for review yet. I want to add another test that calls the endpoint itself and make sure that it returns a valid response.

r-ash:
Going to tidy this PR up and split it into two PRs
* One for validation of shape file
* One for caching (which we will leave until later)

PR #11, opened 2019-08-22
Iterate validation endpoint response schema

r-ash:
This PR will

* Update returned json from validation endpoint to match updated schema
* Also tidy up some files which have tab spacing instead of spaces
* Also remove the dir of `inst/schema/examples`, these examples aren't being maintained and instead we should use the list of examples at `examples/` which are used for testing the API so will be maintained

Updated examples are
```
{
status: success,
errors: [],
data: {
  filename: malawi.pjnz,
  data: { country: Malawi }
}
}
```

```
{
status: success,
errors: [],
data: {
  filename: survey.csv,
  data: [ <survey data> ] 
}
}
```

```
{
status: success,
errors: [],
data: {
  filename: shapefile.zip,
  data: <geojson>
}
}
```


inst/schema/FileName.schema.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": ["string"]
```
richfitz: I don't think that this wants square brackets?

inst/schema/Response.schema.json
```diff
@@ -8,7 +8,7 @@
           "const": "success"
         },
         "data": {
-          "oneOf": [
+          "anyOf": [
```
richfitz: Is `oneOf` not more appropriate here as we want to match exactly one?

* `anyOf`:

> An instance validates successfully against this keyword if it validates successfully against at least one schema defined by this keyword's value.

* `oneOf`:

> An instance validates successfully against this keyword if it validates successfully against exactly one schema defined by this keyword's value.

https://tools.ietf.org/id/draft-handrews-json-schema-validation-01.html#rfc.section.6.7.2

inst/schema/ValidateInputResponse.schema.json
```diff
@@ -1,3 +1,14 @@
 {
-	"type": "string"
+  "type": "object",
```
richfitz: I don't love this - can we include a "role" field or similar in here too perhaps, being the "type" of the file (i.e., an enum of `pjnz`, `shapefile`, etc) and then a separate subschema for each?  You could enforce shared schema with an `allOf` and adding in

```
{
  "type": "object",
  "properties": {
    "filename": { "$ref": "FileName.schema.json" },
    "role": {
      "type": "string",
    }
    "data": {}
  },
  "additionalProperties": false,
  "required": [ "filename", "role", "data" ]
}
```

Then a schema for each type

```
{
  "type": "object",
  "properties": {
    "filename": {},
    "role": {
      "const": "pjnz"
    }
    "data": {
      "$ref": "PjnzResponseData.schema.json"
    }
  },
  "additionalProperties": false,
  "required": [ "filename", "role", "data" ]
}
```

The problem with the current approach is that we'd never want to accidentally get a pjnz response for a shapefile validation for example (not that that would happen)

r-ash:
This be ready for another look now, have fixed the issues mentioned. I'll do the work to update docker image tag we talked about in a separate PR.

inst/schema/ValidateInputResponse.schema.json
```diff
@@ -1,3 +1,52 @@
...
+        { "$ref": "#/definitions/input_response" },
+        { "properties": 
+          {
+            "filename": {},
+            "type": { 
+              "const": "pjnz"
+            },
+            "data": { "$ref": "PjnzResponseData.schema.json" }
+          }
+        }
```
hillalex: Could you add the `required` and `additionalProperties` fields to each of the specific responses also? They don't inherit from the `input_response` unfortunately (or at least the package I'm using the generate typescript interfaces doesn't take them as inherited)

inst/schema/ValidateInputResponse.schema.json
```diff
@@ -1,3 +1,52 @@
...
+    },
+
+    "pjnz_response": {
+      "allOf": [
+        { "$ref": "#/definitions/input_response" },
+        { "properties": 
+          {
+            "filename": {},
+            "type": { 
+              "const": "pjnz"
```
hillalex: Do you think we could change these `const` values to enums with single values instead? It's just that I haven't found a typescript package that supports draft 6 yet :( 

According to the spec "const is merely syntactic sugar for an enum with a single element" https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

richfitz: Can you not use `ajv` from typescript? That's what we're using and it looks like it has typescript support?

inst/schema/ValidateInputResponse.schema.json
richfitz: That seems like a big problem with the validation library, what is it able to get from the allOf?

inst/schema/ValidateInputResponse.schema.json
hillalex: This isn't for validation, it's for actually generating typescript types/interfaces from the schema. I'm using https://www.npmjs.com/package/json-schema-to-typescript.

The only other one I've found is https://github.com/quicktype/quicktype, I can try with that one.

inst/schema/ValidateInputResponse.schema.json
hillalex: This isn't for validation, it's for actually generating typescript types/interfaces from the schema. I'm using https://www.npmjs.com/package/json-schema-to-typescript.

Is it definitely the case that `addiitonalProperties` and `required` should be inherited from `allOf`? That doesn't seem intuitively correct to me and I can't see anything here to suggest it's the case https://json-schema.org/understanding-json-schema/reference/combining.html#allof

inst/schema/ValidateInputResponse.schema.json
hillalex: Actually I was inferring from this issue that the latter will not support `const` either: https://github.com/quicktype/quicktype/issues/1225

richfitz: Sorry, hadn't realised this was for generating things. Rob can easily change this over to being a single-value enum (we talked about this already in person)

inst/schema/ValidateInputResponse.schema.json
richfitz: From the point of view of validation it will be the case:

* schema A says it has x, y, z and nothing else
* schema B describes the contents of z
* schema C is `allOf` A and B

if the data misses x, y or z or if it has an element w it will fail validation against A via C ("must be valid against all of the given subschemas")

But from the point of view of generating interfaces, I guess all bets are off.  I see above that you're specifically talking about interface generation

hillalex: In fact on closer read of the spec C *can't* specify `additionalProperties` or `required` as well, because those would only apply to properties declared in C itself https://json-schema.org/understanding-json-schema/reference/combining.html

So this is a bug in the schema generation that can't be worked around by changing the schema, feel free to ignore and I'll try to find another solution.

inst/schema/ValidateInputResponse.schema.json
hillalex: That'd be great, thanks :+1:

r-ash:
Updates this to now use an `enum` instead of a `const`. It also ensures that we validate using `draft-04` schema.

inst/schema/ValidateInputResponse.schema.json
richfitz: Would something that processed the schemas and do the merging in a way that makes sense make the generator's life easier?

hillalex: Ok, so I'm looking at this again and in fact even though each specific response inherits `allOf` "input_response" they each still re-declare all 3 inherited properties, so actually they could each define "additionalProperties" and "required". 

Also I'm not sure why they re-declare the "filename" property as untyped {} (https://github.com/mrc-ide/hintr/pull/11/files#diff-c52a79711c957ee2a7014047c37bd6c6R37) rather than inheriting the typed filename from "input_response" - I think this is incorrect?

The following would generate a correct typescript type:
```
    "shape_response": {
      "allOf": [
        { "$ref": "#/definitions/input_response" },
        { "properties": {
            "type": { 
              "type": "string",
              "enum": [ "shape" ]
            },
            "data": { "$ref": "ShapefileResponseData.schema.json" }
          },
	  "additionalProperties": false,
	  "required": ["type", "data"]
        }
      ]
    }
```

r-ash: Looks good, I'll fix that.

hillalex: What work is "input_response" actually doing here, given that the specific responses re-declare all 3 of its properties?

richfitz: After redeclaring everything, very little - it was there to make one source of truth for things.  The untyped declaration is fine (for the schema) as it is basically saying it's not the responsibility of this subschema to define the type.  But if that causes problems with the generation you're using then we can add the duplication in, and then it might make sense to remove the shared type

inst/schema/ValidateInputResponse.schema.json
joelhoisko: Sorry for the necro, but did you eventually find some library that deals with `"const"` properly for code generation?

PR #12, opened 2019-08-22
Expand input endpoint to handle shapefile

r-ash:
This PR will

* Allow upload of shapefile to input endpoint
* Validate shapefile is valid and return the serialised data or an error

This builds on mrc-426 so we should only merge this after that once has been merged into master.


R/endpoints.R
```diff
@@ -49,7 +59,7 @@ endpoint_validate_input <- function(req, res, type, path) {
 #'
 #' @return Formatted hintr response.
 #' @keywords internal
-hintr_response <- function(value, schema) {
+hintr_response <- function(value, schema, auto_unbox = FALSE) {
```
r-ash: I wonder if you have any good ideas for a better way to manage this? So I can't rely on wrapping every single value in `scalar` call in the json. 

Part of the problem is we get a file of json. Read it in which converts it to a list which we then validate. It then gets written back into json when `hintr_response` is called. Perhaps it is better to use the raw json after it has been validated and add an alternative `hintr_response` function for when the data is already in json format.

inst/schema/ShapeResponseData.schema.json
```diff
@@ -0,0 +1,15 @@
...
+  "properties": {
+    "type": { "const": "FeatureCollection" },
+    "crs": { "type": "object" },
+    "name": { "tyep": "string" },
+    "features": {
+      "type": "array",
+      "items": { "type": "object" }
+    }
+  }
+}
```
r-ash: What's the best way to validate the GeoJSON? Just put this in as a placeholder. It would be cool if we could validate the GeoJSON against an online schema instead of creating a local copy.

.travis.yml
```diff
@@ -8,6 +8,10 @@ addons:
       - libhiredis-dev
       - redis-server
       - libv8-dev
+      - libjq-dev
+      - libudunits2-dev
+      - libprotobuf-dev
+      - protobuf-compiler
```
r-ash: Also after now seeing the amount of installs required by `geojsonio` I'm wondering if there is a better package to use?

richfitz: All geo stuff is awful for dependencies - most of the involved depends here are probably for the shapefile processing rather than the geojson part. Nothing obviously jumps out as a better bet

R/endpoints.R
richfitz: I don't remember if there's an easy way of interpolating json, but I don't think we should be roundtripping it if we don't need to

PR #13, opened 2019-08-27
Update docker build to create master branch image and latest branch

r-ash:
This PR will
* Update docker push to push an image tagged as "master" to docker hub and push the same image tagged as "latest"

### DIFF
--- a/docker/push
+++ b/docker/push
@@ -7,5 +7,6 @@ docker push $TAG_BRANCH
 
 if [ $GIT_BRANCH == "master" ]; then
    TAG_LATEST="${PACKAGE_ORG}/${PACKAGE_NAME}:latest"
+   docker tag $TAG_BRANCH $TAG_LATEST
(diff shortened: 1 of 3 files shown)


docker/push
```diff
@@ -4,3 +4,8 @@ HERE=$(dirname $0)
 . $HERE/common
 docker push $TAG_SHA
 docker push $TAG_BRANCH
+
+if [ $GIT_BRANCH == "master" ]; then
+   TAG_LATEST="${PACKAGE_ORG}/${PACKAGE_NAME}:latest"
+   docker push $TAG_LATEST
```
hillalex: missing a `docker tag $TAG_SHA $TAG_LATEST`
https://docs.docker.com/engine/reference/commandline/tag/

r-ash: Thanks, makes more sense what this is doing now! I've fixed that.

PR #14, opened 2019-08-27
Add population file upload validation

r-ash:
This PR will
* Enable users to upload population file for validation and return a response. At the moment the response contains no data as there is nothing we want to show or plot in the UI from the population data at the moment. It is only needed for validation

We should only look to merge this once #12 and #11 have been merged into master.


R/validate_inputs.R
```diff
@@ -68,3 +68,31 @@ assert_area_id_exists <- function(json) {
...
+#' Check that population file can be read.
+#'
+#' @param shape Path to input population file.
+#'
+#' @return An error if invalid.
+#' @keywords internal
+do_validate_population <- function(population) {
+  population <- read_csv(population, header = TRUE)
+  ## Perhaps this kind of assertion should just be checked by json schema?
+  assert_column_names(
```
r-ash: One question here about the validation - this kind of thing can be checked by json schema for the response data. But there is a question of where is it best to put those checks. Feels to me like checking this explicitly here is the way to go because it is the kind of error with message that we would want to return to the front end instead of having them try to interpret json schema failure messages. 

Actually the more I think about it I think this is the right way to go. We should never rely on failing a schema to validate our data as it will be a real pain to parse into useful messages I think. Interested to hear what you think and then I will remove line 82.

R/validate_inputs.R
```diff
@@ -68,3 +68,31 @@ assert_area_id_exists <- function(json) {
...
+    colnames(population),
+    c("iso3", "area_id", "time", "sex", "age_group_id", "source", "population"))
+  NULL
```
r-ash: I don't think there is any data we want to return here. This data is only used for the model run (I don't know of any plot which requires it yet) but we can always return something later down the line.

inst/schema/PopulationResponseData.schema.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "maxItems": 0
```
r-ash: Is this the most appropriate way to assert we're not returning any data? We could use `null` here but there is some problem with the conversion from an R `NULL` into json `null`.

It can be done by passing `null = "null"` to `jsonlite::toJSON`. Which is fine but it means our `errors` section on a success would become null whereas at the moment we're relying on the fact that `jsonlite::toJSON` converts that to an empty object `{}`.

So the options I can see are
* Have errors and population data as an empty object
* Update them both to null in the schema and pass `null = "null"` to `jsonlite::toJSON`
* Pass `null = "null"` to `jsonlite::toJSON` but manually set errors to a named empty list for a success `response` (meaning we don't have to change Response schema - it looks like a bit of a faff to create a named empty list though)
* Pass `null = "null"` to `jsonlite::toJSON` but manually set errors to an empty list and update its type in schema to `array` for a `success` response (this makes it consistent with `failure` response which is already an array)

or something else I haven't thought of.

r-ash:
Also I'm thinking that `inst/testdata/population.csv` is a bit too big for this to sit in the package at 4.9MB. The other datasets we're going to add are a bit smaller at ~ 2.5MB, 88KB, 63KB each.

I'm keen we have realistic test data here (so that we can have something to reference when testing the front end) - i.e. the data is complete for every district/region. And this would then feed into the testing we're going to do of the data together later down the line (when we check the regions in the population data against the regions in the shape file). Do you have any thoughts on that? Maybe having test data of about 10MB in the package total is okay?

PR #17, opened 2019-08-29
MRC-437: Faster builds by not building docker images on travis

richfitz:
This PR will speed up CI by not building docker images on travis.  Instead, these are built on teamcity, and a very light test is included to make sure that the whole lot can be brought up


.travis.yml
```diff
@@ -13,9 +13,6 @@ addons:
       - libv8-dev
       - protobuf-compiler
       - redis-server
-before_script:
-  - ./docker/travis
```
r-ash: Should we delete those scripts too?

richfitz: yeah, good call

PR #19, opened 2019-08-29
Add programme data validation endpoint

r-ash:
This PR will
* Add endpoint for validating and returning programme data
* Moves validation asserts into a separate file from the `do_validation_*` functions

How specific do we want the schema for the returned data to be? Here I have asserted that is has to return an array of objects with at least `iso3` and an `area_id` but really that was some arbitrary decision. Perhaps we should assert that it returns enough to do the plotting the front end wants? Or let the front end just not render a plot if the data isn't available and log the fact in the back end?


R/json.R
```diff
@@ -40,3 +40,11 @@ to_upper_first <- function(x) {
   substr(x, 1, 1) <- toupper(substr(x, 1, 1))
   x
 }
+
+## Read json and apply a class so we can use method dispatching for json
+hintr_geojson_read <- function(path) {
+  json <- geojsonio::geojson_read(path, method = "local")
```
richfitz: Is this correct? Does `geojsonio::geojson_read` not return an R object? In which case pretending that it's json must do bad things if we return that anywhere.  Looking here specifically at adding the `json` class.

R/validate_inputs.R
```diff
@@ -31,68 +31,45 @@ do_validate_shape <- function(shape) {
...
 #'
-#' @param shape Path to input population file.
+#' @param population Path to input population file.
 #'
 #' @return An error if invalid.
 #' @keywords internal
 do_validate_population <- function(population) {
   population <- read_csv(population, header = TRUE)
-  ## Perhaps this kind of assertion should just be checked by json schema?
+  assert_single_country(population, "population")
```
richfitz: This could be a nice place to use the defer package?

R/json.R
r-ash: It just returns a list as far as I can see. I was thinking about that class on the way in - perhaps more appropriate to wrap in a `geojson` class as the assert method which uses it for method dispatch the behaviour specific to `geojson`. Either that or I we can just use a `switch` on the file type instead.

richfitz: Something like `geojson` would be better because `json` is the class we use in `json_verbatim`

r-ash: Fixed this

R/validate_inputs.R
r-ash: I think I will leave this for a separate PR once we get that package on drat.

r-ash:
I think this is ready for another look. Made updates based on Rich's suggestions but will leave using deferrable errors for later as it would be good to get this merged.

PR #21, opened 2019-09-02
mrc-447 Added schema for data filters

EmmaLRussell:


inst/schema/FilterOption.schema.json
```diff
@@ -0,0 +1,19 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "options": {
+      "type": "array",
+      "items": {
+        "$ref": "#"
```
EmmaLRussell: Options can contain other options, allowing for multiple geo levels

inst/schema/ValidateInputRequest.schema.json
```diff
@@ -3,7 +3,8 @@
   "type": "object",
   "properties": {
     "type" : { "$ref": "InputType.schema.json" },
-    "path" : { "$ref": "FilePath.schema.json" }
+    "path" : { "$ref": "FilePath.schema.json" },
+    "filters": { "$ref": "Filters.schema.json" }
```
EmmaLRussell: Not requiring this property, so shouldn't break current API!

inst/schema/NestedFilterOption.schema.json
```diff
@@ -0,0 +1,19 @@
+{
```
EmmaLRussell: Most filter options can be simple string arrays, but this allows for options which contain further options

inst/schema/SurveyResponseData.schema.json
```diff
@@ -0,0 +1,10 @@
+{
```
EmmaLRussell: Added this just as an example response type with filters in, and so I can get some test data being passed back to the front end initially

inst/schema/ANCFilters.schema.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "age": {
```
EmmaLRussell: ANC is all female, so no sex filter required here

inst/schema/SurveyFilters.schema.json
```diff
@@ -0,0 +1,28 @@
...
+      }
+    },
+    "survey": {
+      "type": "array",
+      "items": { "type": "string" }
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "sex", "age", "region"
```
EmmaLRussell: Survey is optional - not returned if file contains only one survey

inst/schema/SurveyFilters.schema.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "sex": {
+      "type": "array",
+      "items": { "type": "string" }
```
hillalex: Will sex not always be "female", "male", "both"?

inst/schema/SurveyFilters.schema.json
```diff
@@ -0,0 +1,28 @@
...
+    },
+    "age": {
+      "type": "array",
+      "items": { "type": "string" }
+    },
+    "region": {
+      "type": "array",
+      "items": {
+        "$ref": "NestedFilterOption.schema.json"
```
hillalex: Will these differ between survey, program and ANC? Or can we just have a master list of regions for a country (if some regions are missing for a particular data set they just appear as grey/no data in the leaflet map)

inst/schema/SurveyFilters.schema.json
EmmaLRussell: I think we can probably enforce that, so yes, in that case, these could be hardcoded. (?)

inst/schema/SurveyFilters.schema.json
EmmaLRussell: Yeah, I was wondering that too, this is potentially quite a bulky data structure to pass around multiple times, just on the offchance that one might contain a slightly reduced list. We're expecting the norm to be that all files have data from all areas (?). So if we did just have one master list, maybe it should be the shape file validate that returns that?

hillalex: Yeah, I feel like returning this with the shapefile would make sense. In the case of missing data regions will just show as greyed out on the map, can have a legend to indicate that means "no data".

inst/schema/NestedFilterOption.schema.json
```diff
@@ -8,7 +8,7 @@
     "options": {
       "type": "array",
       "items": {
-        "$ref": "#"
+        "type": "object"
```
EmmaLRussell: Annoyingly the generateTypeScript task in hint falls over with circular references in the json schema, whatever option you choose for supposedly dealing with them.

PR #24, opened 2019-09-04
Add endpoint for validating survey data

r-ash:
This PR will

* Add endpoint for validating and returning survey data

This follows the same pattern as we have used for ANC and programme data.


tests/testthat/test-validate-inputs.R
```diff
@@ -37,10 +37,18 @@ test_that("do_validate_programme validates programme file", {
...
   anc <- file.path("testdata", "anc.csv")
-  data <- do_validate_programme(anc)
+  data <- do_validate_anc(anc)
   ## Some arbitrary test that the data has actually been returned
   expect_true(nrow(data) > 800)
   expect_equal(typeof(data$value), "double")
 })
+
+test_that("do_validate_survey validates survey file", {
+  anc <- file.path("testdata", "survey.csv")
```
hillalex: typo - var should be called "survey"

R/validate_inputs.R
```diff
@@ -90,3 +90,21 @@ do_validate_anc <- function(anc) {
...
+#'
+#' Check that survey data file can be read and return serialised data.
+#'
+#' @param anc Path to input survey file.
+#'
+#' @return An error if invalid.
+#' @keywords internal
+do_validate_survey <- function(survey) {
+  survey <- read_csv(survey, header = TRUE)
+  assert_single_country(survey, "anc")
```
hillalex: "anc" should be "survey"

hillalex: Not sure which level this makes most sense at but we should be testing the unhappy path as well - i.e. testing that bad survey data fails the validation. I can't see that happening anywhere? Such a test would catch mistakes like this that only pertain to the error message returned!

tests/testthat/test-validate-inputs.R
r-ash: Thanks, fixed that.

R/validate_inputs.R
r-ash: Fixed. Yeah I should have added that. There was one in for the programme but I had forgotten for anc and survey data. I've added a test for each now - though they only check 1 of the failure conditions. I think probably I need to add some helper code to setup the malformed test data to avoid having a proliferation of test data within the package otherwise we're going to end up with 1 file for every possible type of error for each input type.

PR #29, opened 2019-09-12
Return filter information from R API

r-ash:
This PR will

* Add data to validate requests about possible filters for the UI for the programme, ANC and survey files

(shape file filters I will cover in a separate PR)


inst/schema/NestedFilterOption.schema.json
```diff
@@ -15,8 +15,5 @@
       }
     }
   },
-  "additionalProperties": false,
```
EmmaLRussell: Why remove this required list?

EmmaLRussell: Why remove this required list?

r-ash: An error, reverted that now.

r-ash:
Good point about the order - for age groups that is associated with metadata which maps age group id to the age group label. So I think I should return this from the UI in the correct order (however how does that work when we're combining sets? i.e. if programme has age groups (0-4, 5-9, 10-14, 15+) and ANC comes with age groups (0-14, 15+) how do we resolve the ordering of the union of those sets. We may have to return the display order as a property of the filter so you can combine them on your end. 

I think the R API should manage and return the order though, will ask Jeff if he can add a column to his metadata function in naomi which I'm calling from here.

EmmaLRussell:
> Good point about the order - for age groups that is associated with metadata which maps age group id to the age group label. So I think I should return this from the UI in the correct order (however how does that work when we're combining sets? i.e. if programme has age groups (0-4, 5-9, 10-14, 15+) and ANC comes with age groups (0-14, 15+) how do we resolve the ordering of the union of those sets. We may have to return the display order as a property of the filter so you can combine them on your end.
> 
> I think the R API should manage and return the order though, will ask Jeff if he can add a column to his metadata function in naomi which I'm calling from here.

From the point of view of the UI, the order of combined filter sets doesn't matter, as I think we'll only be showing a single set from which the user can select filter values (that for the currently selected data set) - and at the moment, the selected filters are shown in the order in which the user selected them.

inst/schema/NestedFilterOption.schema.json
EmmaLRussell: Can't see a commit for that...?

r-ash: Oops, I failed to push it.

r-ash:
I've added some sorting of the returned filters. The order of the filters in the UI should be the same as the order the filters are returned from the R side. Still a TODO to add that sorting for the age filters but that is blocked on Jeff adding that column to his data helper function. I will add a ticket for that but I think this should be ready for another look now.

PR #30, opened 2019-09-16
Validate survey, ANC and programme data regions against shape file regions

r-ash:
This PR will

* Split out the baseline inputs validation from the survey and programme data validation
* Update survey and programme validation to also take a shape file and validate the regions against the shape file


R/validate_inputs.R
```diff
@@ -116,14 +119,17 @@ do_validate_population <- function(population) {
...
+  data <- read_csv(programme, header = TRUE)
+  assert_single_country(data, "programme")
   assert_column_names(
-    colnames(programme),
+    colnames(data),
     c("iso3", "area_id", "period", "sex", "age_group_id", "indicator", "value"))
-  list(data = programme,
-       filters = list("age" = get_age_filters(programme)))
+  assert_consistent_regions(read_regions(shape, "shape"),
+                            read_regions(programme, "programme"),
```
r-ash: One question is about this, we pass the path to the programme data into the function not the already read data. It means we're reading it twice, once on line 123 and once on line 129. How bad is that? Does R do something clever and not duplicate the file read or should I be handling that?

r-ash:
For the front end this will

* No longer allow users to call `validate/input` for survey, programme or ANC data
* Expose `validate/survey_and_programme` for survey, programme and ANC data instead
* Require path to `shape` file is passed to requests to `validate/survey_and_programme`

R/validate_inputs.R
richfitz: I think it might be best to ignore this for now and we can look at general caching once we see the general lay of things

R/endpoints.R
```diff
@@ -4,6 +4,8 @@ api_build <- function(queue) {
             serializer = serializer_json_hintr())
   pr$handle("POST", "/validate/baseline", endpoint_validate_baseline,
             serializer = serializer_json_hintr())
+  pr$handle("POST", "/validate/survey_and_programme", endpoint_validate_survey_programme,
```
richfitz: Previously/elsewhere we had decided that dashes were REST API best practice

R/endpoints.R
```diff
@@ -118,6 +117,41 @@ endpoint_validate_input <- function(req, res, type, path) {
...
+#'
+#' @return Validated JSON response with data and incidcation of success.
+#' @keywords internal
+endpoint_validate_survey_programme <- function(req, res, type, path, shape) {
+  validate_json_schema(req$postBody, "ValidateSurveyAndProgrammeRequest")
+  validate_func <- switch(type,
+                          programme = do_validate_programme,
+                          anc = do_validate_anc,
+                          survey = do_validate_survey)
+  validate_if_exists <- function(path, shape) {
```
richfitz: why does this function shadow the package-level function? is there a simpler way of containing this logic?

R/utils.R
```diff
@@ -18,3 +18,25 @@ is_empty <- function(x) {
...
+#' Collapse a vector into a human readable format for returning in messages
+#'
+#' @param vector The vector to collapse
+#' @param collapse Separator between each element after collapse
+#' @param limit Max number of items from vector to return in string
+#' @param end If length of vector over the max a string to append to the end
+#'
+#' @return The collapsed vector as a string.
+#' @keywords internal
+collapse <- function(vector, collapse = ", ", limit = 10, end = ", ...") {
```
richfitz: This is cute :)

Is it worth truncating at a total character width instead?  If you do `cumsum(nchar(vector))` and add on the appropriate separator you'll see how far out you go

R/endpoints.R
r-ash: Yep, had forgotten about that. That's updated now.

R/endpoints.R
r-ash: Fixed that now to just check for file existing within the `with_success` call.

R/utils.R
r-ash: That is better, updated that too.

r-ash:
I think this is ready for another look now, I've tried to make the intention a bit clearer in the news item.

We should probably also not merge this until Alex or Emma have had a chance to take a look as it may break their code on the front end. I think they are just running against master.

PR #31, opened 2019-09-18
mrc-475 testdata update

r-ash:
This PR will

* Update testdata to data expected for users to upload and fix tests
* Update build scripts to include packages required by naomi

In particular this updates the format of the data and fixes any validation to support that. The `.geojson` update is required for returning the set of region filters.
The only anticipated change to the format of this data is to replace `age_group_id` to return human readable values for ID instead of numeric values so it is easier for the users uploading the data.


.travis.yml
```diff
@@ -18,12 +18,15 @@ services:
 before_script:
   - R CMD INSTALL .
 cache: packages
+repos:
```
richfitz: do you want to try using `Additional_repositories` in `DESCRIPTION` for this? I'm not sure I see how this is being used as-is in the travis build as it's not listed as a dependency (though I do see it in the dockerfile)

tests/testthat/test-endpoints-validate-survey-and-programme.R
```diff
@@ -57,7 +57,10 @@ test_that("endpoint_validate_survey_programme supports ANC file", {
   expect_equal(res$status, 200)
   ## Sanity check that data has been returned
   expect_true(length(response$data$data) >= 800)
-  expect_equal(typeof(response$data$data[[1]]$value), "double")
+  ## TODO: Add test to ensure we are returning prevalence & art coverage are
```
richfitz: Does this have a corrsponding ticket?

R/validate_inputs.R
```diff
@@ -145,10 +145,14 @@ do_validate_anc <- function(anc, shape) {
...
   assert_column_names(
     colnames(data),
-    c("iso3", "area_id", "period", "sex", "age_group_id", "indicator", "value"))
+    c("area_id", "age_group_id", "quarter_id", "anc_clients",
+      "ancrt_hiv_status", "ancrt_known_pos", "ancrt_already_art",
+      "ancrt_tested", "ancrt_test_pos"))
   assert_consistent_regions(read_regions(shape, "shape"),
                             read_regions(anc, "anc"),
                             "ANC")
+  ## TODO: Call naomi to calculate prevalence & art_coverage which is what we
```
richfitz: Does this have a corresponding ticket?

tests/testthat/test-endpoints-validate-survey-and-programme.R
r-ash: Added the ticket no to the comment.

R/validate_inputs.R
r-ash: Added the ticket no to the comment.

.travis.yml
r-ash: So this is only in there because it is requried by naomi and we're including naomi in the build. I think there isn't a way for naomi to automatically install that for us?

r-ash:
Added a README about the testdata now. Mostly this is coming from Jeff - though I don't know what the source of that is. Let me know if you think I should chase that up further. I'm not expecting any updated outside of known ones which are coming which I've listed in the README. 

I've also added task ids to the TODO tasks.

.travis.yml
richfitz: If naomi has an additional repos listed in *its* DESCRIPTION I believe it will work ok

r-ash: I'll make a note to update that tomorrow - I was handling it through travis like I was here.

PR #32, opened 2019-09-19
Return region filter information from R API

r-ash:
This PR will

* Return region filter hierarchy from geojson
* Also returns some metadata about region level and label which front end isn't currently using but will need at some point

Looks like the INLA build is working on this branch so not sure what happened on master


R/utils.R
```diff
@@ -44,3 +44,9 @@ collapse <- function(vector, collapse = ", ", limit = 150, end = "...") {
   out <- paste(vector, collapse = collapse)
   out
 }
+
+`%||%` <- function(a, b) if (is.null(a)) b else a
```
richfitz: could you reformat this (this is my fault)

tests/testthat/test-filters.R
```diff
@@ -54,3 +54,100 @@ test_that("get_survey_filters gets available filter options and sorts them", {
   expect_equal(get_survey_filters(NULL), list())
   expect_equal(get_age_filters(data.frame(survey_id = NULL)), list())
 })
+
+test_that("can construct sorted tree from data frame", {
+  data <- data.frame(
+    id = c("MWI", "MWI.1", "MWI.2", "MWI.1.1", "MWI.1.2"),
+    parent_id = c(NA, "MWI", "MWI", "MWI.1", "MWI.1"),
+    sort_order = c(1, 2, 3, 4, 5),
+    stringsAsFactors = FALSE
```
richfitz: you could use `data_frame` and reduce noise here

inst/schema/ValidateInputResponse.schema.json
```diff
@@ -25,9 +25,13 @@
...
-          "properties" :{
-            "regions": {
+          "properties" : {
+            "level_labels": {
               "type": "array",
+              "items": { "$ref": "LevelLabels.schema.json"}
+            },
+            "regions": {
+              "type": "object",
               "items": { "$ref": "NestedFilterOption.schema.json" }
```
richfitz: It looks like this schema does not match the data:

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    },
    "id": {
      "type": "string"
    },
    "options": {
      "type": "array",
      "items": {
        "type": "object"
      }
    }
  },
  "additionalProperties": false,
  "required": [
    "name", "id"
  ]
}
```

it should be recursive, right?

r-ash: Fixed that.

tests/testthat/test-filters.R
r-ash: Nice - have fixed that throughout.

R/utils.R
r-ash: Yeah sorry, I had said I would do that and then subsequently forgot about it. Tidied up now.

PR #33, opened 2019-09-20
Return realistic model run data and update tests

r-ash:
This PR will

* Return realistic model data result read from csv
* Parse out filters from the result for front end


R/endpoints.R
```diff
@@ -69,9 +68,7 @@ endpoint_model_result <- function(queue) {
       )
       response$value <- NULL
       res$status <- 400
-    } else if (response$success) {
-      response$value <- scalar(response$value)
-    } else {
+    } else if (!response$success) {
```
richfitz: I know it's not in this PR but I am surprised that the opening bit here is surprising that we're checking for an error class and not a success?

tests/testthat/test-queue.R
```diff
@@ -32,7 +32,7 @@ test_that("queue works as intended", {
   expect_equal(status$queue, 0)
 
   ## After task has completed
-  Sys.sleep(1)
+  Sys.sleep(2)
```
richfitz: You can use task_wait on the underlying rrq object here if you want to wait for less time and make the test less fragile, and we could reduce the sleep time further as that's not really needed here?

PR #38, opened 2019-09-24
Get prevalence and ART coverage for uploaded ANC data

r-ash:
This PR will

* Return prevalence and ART coverage from uploaded ANC data for use in plots

### DIFF
--- a/R/validate_inputs.R
+++ b/R/validate_inputs.R
@@ -152,9 +152,7 @@ do_validate_anc <- function(anc, shape) {
   assert_consistent_regions(read_regions(shape, "shape"),
                             read_regions(anc, "anc"),
                             "ANC")
-  ## TODO: Call naomi to calculate prevalence & art_coverage which is what we
-  ## need to return to the front end for plotting. mrc-492
-  list(data = data,
+  list(data = naomi::calculate_prevalence_art_coverage(data),
        filters = list("age" = get_age_filters(data)))
 }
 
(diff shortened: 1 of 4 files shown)


tests/testthat/test-validate-inputs.R
```diff
@@ -86,13 +86,15 @@ test_that("do_validate_programme validates programme file", {
   expect_equal(data$filters$age, expected_filters)
 })
 
-test_that("do_validate_anc validates ANC file", {
+test_that("do_validate_anc validates ANC file and gets data for plotting", {
```
richfitz: Here you test for existance of `prevalance` and `art_coverage` but in the schema only for `area_id` (`AncResponseData.schema.json`). Should these be harmonised?

R/validate_inputs.R
```diff
@@ -152,9 +152,7 @@ do_validate_anc <- function(anc, shape) {
   assert_consistent_regions(read_regions(shape, "shape"),
                             read_regions(anc, "anc"),
                             "ANC")
-  ## TODO: Call naomi to calculate prevalence & art_coverage which is what we
-  ## need to return to the front end for plotting. mrc-492
-  list(data = data,
+  list(data = naomi::calculate_prevalence_art_coverage(data),
```
richfitz: Please update the DESCRIPTION to depend on `naomi (>= 0.0.2)`

tests/testthat/test-validate-inputs.R
r-ash: Yeah they should, I've updated that now.

R/validate_inputs.R
r-ash: Thanks, updated that.

PR #39, opened 2019-09-25
expect original filename and return it

hillalex:
Files are now saved by their hash, but we want to display the original filename to the user in the front-end, so we need it back in the serialised data. This PR changes the schema so the original filename gets sent with the validation request and returned with the validation response. Feels a bit annoying to just send it back and forth like this but can't think of another way, since it's `HINT` that knows the filename but `hintr` that does the serialisation.

### DIFF
--- a/tests/testthat/payload/validate_anc_payload.json
+++ b/tests/testthat/payload/validate_anc_payload.json
@@ -1,5 +1,6 @@
 {
   "type": "anc",
   "path": "testdata/anc.csv",
-  "shape": "testdata/malawi.geojson"
+  "shape": "testdata/malawi.geojson",
+  "originalFilename": "original.geojson"
 }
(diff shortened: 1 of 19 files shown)


tests/testthat/payload/validate_anc_payload.json
```diff
@@ -1,5 +1,6 @@
 {
   "type": "anc",
   "path": "testdata/anc.csv",
-  "shape": "testdata/malawi.geojson"
+  "shape": "testdata/malawi.geojson",
+  "originalFilename": "original.csv"
```
richfitz: It would actually be super useful to have the hash come through as well.  This means that there are 3 pieces of information we're using to describe a file:

* `hash` - its md5, as computed by hint
* `path` - where hint has put the file on the shared volume
* `originalFilename` (or just `filename`) - the filename as uploaded by the user

These 3 pieces could go together in one little schema and be passed around as a dict.

Otherwise when we start on the caching layer the first thing we do will start hashing the file :)

PR #40, opened 2019-09-30
Add endpoint for plotting metadata

r-ash:
This PR will

- Add an endpoint to return metadata to the front end about plotting. Including:
   - Colour and scale to use per country and indicator
   - How to locate & present data from the uploaded data (column containing value to present, optional filtering on another column, display name for data)


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hintr
 Title: R API for calling naomi district level HIV model
-Version: 0.0.4
+Version: 0.0.5
```
richfitz: Please increment the naomi requirement too (you require 0.0.3 now)

R/metadata.R
```diff
@@ -0,0 +1,46 @@
+do_plotting_metadata <- function(country) {
+  metadata <- naomi::get_plotting_metadata(country)
```
richfitz: Lines 3-7 can be replaced with (I believe):

```
lapply(split(metadata, metadata$data_type), build_data_type_metadata)
```

and a similar pattern will probably help simplify the functions below

R/metadata.R
```diff
@@ -0,0 +1,46 @@
...
+  out <- lapply(unique(metadata$indicator), function(indicator) {
+    data <- metadata[metadata$indicator == indicator, ]
+    build_indicator_metadata(data)
+  })
+  names(out) <- unique(metadata$indicator)
+  out
+}
+
+build_indicator_metadata <- function(metadata) {
+  if(nrow(metadata) != 1) {
```
richfitz: missing space after if (sorry)

inst/schema/PlottingMetadataResponse.schema.json
```diff
@@ -0,0 +1,86 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "definitions": {
```
richfitz: it looks like `{survey,anc,output}_chloropleth` refer to the same type, and so you could simplify this

DESCRIPTION
r-ash: Fixed, thanks.

R/metadata.R
r-ash: Good spot

R/metadata.R
r-ash: Thank you! I wanted something like that, so much neater.

inst/schema/PlottingMetadataResponse.schema.json
r-ash: They do, though this is one of the things I expect might change after talking to Jeff. I've simplified it for now though as you suggest.

PR #41, opened 2019-09-30
Return quarter ID filters for programme data

r-ash:
This PR will

* Return quarter_id filter information for programme ART data and ANC data
* Remove age filter from ANC data as we don't expect uploaded ANC data to be age stratified
* Fix example payloads to match updated spec (I missed this when approving previous PR)

Also removed a couple of TODOs which I notice I had not removed when I implemented the feature


r-ash:
> This must mean that we're not validating those example payloads anywhere - can you open a ticket to address that?

Thanks, added a ticket mrc-537

PR #42, opened 2019-10-03
MRC-542: Get hintr working with new rrq

richfitz:
This PR will use the new context-free rrq version. https://github.com/mrc-ide/rrq/pull/9 should be merged and references to mrc-538 should be removed from this PR before it is merged.

It would be nice to test these workers properly in the test script, and to stop using the spawn at all, but these should wait until future PRs, and they will impact on the deployment of hint. See also https://vimc.myjetbrains.com/youtrack/issue/mrc-544 which would make testing much easier.


richfitz:
Note that the tests are not passing the second time around under codecov at the moment :(

PR #46, opened 2019-10-14
Make input & output response schemas consistent

r-ash:
This PR will

* Fix plotting metadata response for output data
* Add indicator filters to input data responses

Expecting this to be read as
Plotting metadata describes the full list of possible indicators which may be returned from input or output data, how to locate them within that dataset and what colours and scales we should use for these
The indicator filters describe the actual indicators (as a strict subset of possible indicators) which the input or output data actually contains

This also removes the pin on a specific version of naomi


r-ash:
> Schema looks good to me. I'm seeing a typo in the actual plotting metadata returned for survey vls - "inerpolateWarm" instead of "interpolateWarm".

Thanks, fixed that now

EmmaLRussell:
> > Schema looks good to me. I'm seeing a typo in the actual plotting metadata returned for survey vls - "inerpolateWarm" instead of "interpolateWarm".
> 
> Thanks, fixed that now

Sorry, there's also "inerpolatePlasma" for recent..

r-ash:
> > > Schema looks good to me. I'm seeing a typo in the actual plotting metadata returned for survey vls - "inerpolateWarm" instead of "interpolateWarm".
> > 
> > 
> > Thanks, fixed that now
> 
> Sorry, there's also "inerpolatePlasma" for recent..

I think I must have a dodgy t key - I've fixed that one too.

PR #47, opened 2019-10-14
form schema and example

hillalex:
The example JSON here results in a form that looks like:
Screenshot from 2019-10-14 15-32-59

Hopefully it's clear from this image and the JSON what the different properties correspond to.

NB at the moment only number and select controls are supported. Is this all we need for Naomi?

### DIFF
--- a/inst/schema/FormOptions.schema.json
+++ b/inst/schema/FormOptions.schema.json
@@ -1,0 +1,48 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "definitions": {
+    "selectControl": {
+      "properties": {
+        "name": { "type": "string" },
+        "label": { "type": "string" },
+        "type": { "enum": [ "select", "multiselect" ] },
+        "required": { "type": "boolean" },
+        "default": { "type": "string" },
+        "helpText": { "type": "string" },
+        "options": { "type":  "array", "items": { "type": "string" }}
+      },
+      "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


hillalex:
:woman_facepalming:  thanks

inst/schema/FormOptions.schema.json
```diff
@@ -0,0 +1,55 @@
...
+  "definitions": {
+    "selectControl": {
+      "properties": {
+        "name": { "type": "string" },
+        "label": { "type": "string" },
+        "type": { "enum": [ "select", "multiselect" ] },
+        "required": { "type": "boolean" },
+        "default": { "type": "string" },
+        "helpText": { "type": "string" },
+        "options": { "type":  "array", "items": { "type": "string" }}
```
hillalex: Will we in fact need: 
* separate id and human readable name for options?
* nested options?

r-ash: Separate ID and human readable names - quite possibly this will depend on how the model is running but perhaps for safety we should do this i.e. does the model use quarter as ID or as the human readable text
Nested options - This could have some use in the region selector. In mrc-570 I've just returned a flat list of regions for the time being we could move this to a hierarchy and use a similar widget to region filters

r-ash:
As we just discussed, I've put this schema into `naomi` see https://github.com/mrc-ide/naomi/pull/13 so will close PR here.

PR #53, opened 2019-10-17
mrc-618 update model run interface

hillalex:
A model run request will now include 2 arbitrary dictionaries of options. This PR just updates the schema and the skeleton methods to reflect this new interface.


R/endpoints.R
```diff
@@ -64,9 +64,9 @@ endpoint_model_options <- function(options_template) {
 }
 
 endpoint_model_submit <- function(queue) {
-  function(req, res, data, parameters) {
+  function(req, res, data, options, advanced) {
```
hillalex: Ooh, unless..... we just group these all together into a single "options" dictionary. Would make the interface simpler. @r-ash any thoughts on whether model options and advanced options should be distinguished in this way?

r-ash: I don't mind - it will be fairly straight forward to use those if they are together or separate.

I agree makes the interface a little simpler but either is fine from my point of view.

hillalex: I think it's more flexible to group them actually. That way if, say we didn't end up having any advanced options, or if we ended up having 3 sets of options, the schema/interface remains unchanged.

r-ash: Yeah great point - let's group them

hillalex: Ok, have updated accordingly.

tests/testthat/test-endpoints-model.R
```diff
@@ -47,8 +39,9 @@ test_that("endpoint model run queues a model run", {
   ## Call the endpoint
   queue <- Queue$new()
   model_submit <- endpoint_model_submit(queue)
-  response <- model_submit(req, res, data, parameters)
+  response <- model_submit(req, res, data, options)
   response <- jsonlite::parse_json(response)
+  str(response$errors)
```
r-ash: This a line left over from debugging?

hillalex: oops, good spot :)

hillalex:
> This looks good - just one question about the `str` line left in. Happy to merge after that.
> 
> Did we decide what the model run options will look like should users select multiple?

In my latest branch I've updated the form so that it returns arrays of strings for inputs of type "multiselect". Note that all other values will be strings (ie. numeric inputs will return as strings, "10", "0.1", etc.)

PR #57, opened 2019-10-28
tiny schema change for form options

hillalex:
I don't think this is in use anywhere, but "default" should change to "value".


r-ash:
I'm going to close this PR as I've pulled these changes into #59 and fixed subsequent test failures there.

PR #58, opened 2019-10-29
mrc-633 Schema for Barchart metadata

EmmaLRussell:


inst/schema/BarchartMetadata.schema.json
```diff
@@ -0,0 +1,16 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties" : {
+    "indicators" : {
+      "type": "array",
+      "items": { "$ref": "BarchartIndicator.schema.json" }
+    },
+    "filters" : {
+      "type": "array",
+      "items": { "$ref": "Filter.schema.json" }
```
r-ash: I wonder if just calling this `Filter` is misleading because really this is a specific Filter? I.e. we have SurveyFilters, ModelResultFilters etc. already. I think either we make `Filter` more generic schema and use it throughout or we call this `BarchartFilters`.

PR #62, opened 2019-10-31
MRC-664: Use file objects everywhere

richfitz:
This PR will update functions to pass around file objects (path, hash, filename) rather than raw paths everywhere, referencing files at the last minute.

This is required to support caching (e.g. mrc-658)


richfitz:
Coverage is fixed with master merged in now.

PR #64, opened 2019-11-01
Return model run status to front end

r-ash:
This PR will

* Get the progress from the queue and return it to the front end as part of the `model/status` endpoint

### DIFF
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -87,6 +87,7 @@ endpoint_model_status <- function(queue) {
     response <- with_success(
       queue$status(id))
     if (response$success) {
+      response$value <- progress()
       response$value <- lapply(response$value, scalar)
       response$value$id <- scalar(id)
     } else {
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -370,7 +370,7 @@ prepare_status_response <- function(value, id) {
     }
   }
 
-  response_value <- lapply(value[-which(names(value) == "progress")], scalar)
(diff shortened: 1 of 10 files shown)


R/endpoints.R
```diff
@@ -364,6 +361,21 @@ api_root <- function() {
...
+prepare_status_response <- function(value, id) {
+  set_scalar <- function(x) {
+    if (length(names(x)) > 1) {
+      lapply(x, set_scalar)
+    } else {
+      scalar(x)
+    }
+  }
+
+  response_value <- lapply(value[-which(names(value) == "progress")], scalar)
```
richfitz:
```
lapply(value[names(value) != "progress"], scalar)
```
perhaps?

inst/schema/ProgressPhase.schema.json
```diff
@@ -0,0 +1,13 @@
+{
```
richfitz: we should think about where to put the schema for this if we're going to use it elsewhere more commonly - this one corresponds to what we discussed on slack?

r-ash: Yeah this is the general one. Expecting a standard Progress message to be an array of these.

R/endpoints.R
r-ash: Much nicer thank you. I thought what I wrote was painfully complicated for what I was trying to achieve. I think I just had a confused moment and forgot `!=` existed.

PR #68, opened 2019-11-11
Return metadata with model run result response

r-ash:
This PR will

* Return metdata about barcharts with model run response

This is a shift slightly in how we have been returning plotting metadata before. We decided that the metadata about plotting, so how to locate indicators and filters should be returned with the model run response. So merging this PR will leave it in a state where the input metadata should be read from the previous setup (i.e. from a separate plottingMetadata endpoint) but the output metadata should be read from the model run result response.

We should merge mrc-681 in naomi first https://github.com/mrc-ide/naomi/pull/24 this will make the build pass


R/filters.R
```diff
@@ -94,9 +94,20 @@ read_long_indicator_filters <- function(data, type) {
 }
 
 get_model_output_filters <- function(data) {
-  list(age = get_age_filters(data),
-       quarter = get_quarter_filters(data),
-       indicators = get_id_label_map(data, "indicator_id"))
+  list(
+    list(
+      id = scalar("age"),
+      column_id = scalar("age_group_id"),
```
r-ash: @hillalex Could you review this section?

I'm interpreting the meaning of

id - a machine readable id for the filter does this need to match anything in particular?
column_id - the name of the column to filter on
label - human readable label for the filter
options - label, id pairs for filters where label is thing to be shown to the user and the id matches the id of one of the values within the column `column_id`.

Is that right?

hillalex: Yes that's exactly right. The filter id is something we use on our side to identify filters internally, does not need to match anything.

PR #72, opened 2019-11-14
Support HEAD requests for download endpoints

r-ash:
This PR will

* Add HEAD endpoints for summary and spectrum downloads

PR #73, opened 2019-11-14
Add content-disposition header to download endpoints

r-ash:
This PR will

* add `Content-Disposition: attachment; filename="filename.zip"` header to download endpoints

We should merge this after #72 has been looked at as it builds on that

### DIFF
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -17,13 +17,13 @@ api_build <- function(queue) {
   pr$handle("GET", "/meta/plotting/<iso3>", endpoint_plotting_metadata,
             serializer = serializer_json_hintr())
   pr$handle("GET", "/download/spectrum/<id>", endpoint_download_spectrum(queue),
-            serializer = serializer_zip("naomi_spectrum_digest.zip"))
+            serializer = serializer_zip("naomi_spectrum_digest"))
   pr$handle("HEAD", "/download/spectrum/<id>", endpoint_download_spectrum(queue),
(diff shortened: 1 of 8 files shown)


R/endpoints.R
```diff
@@ -17,13 +17,13 @@ api_build <- function(queue) {
   pr$handle("GET", "/meta/plotting/<iso3>", endpoint_plotting_metadata,
             serializer = serializer_json_hintr())
   pr$handle("GET", "/download/spectrum/<id>", endpoint_download_spectrum(queue),
-            serializer = serializer_zip())
```
richfitz: should we always be returning the same filename here? or should it depend on the id? with the same filename in the odin ui, users ended up confused with an ever-growing number of filenames ending in `(1)`, `(2)` etc

r-ash: Hmm I'm not sure what is good practice here. I've never thought having filenames with a `(1)` and `(2)` was that confusing but I guess it might be. Let's ask the others in slack.

r-ash:
> it looks like the new argument-taking `serializer_zip` is never tested - could you add both a unit and integration test for that please? (see https://codecov.io/gh/mrc-ide/hintr/pull/73/diff)

Thanks yeah - added a unit test for that now. There are checks in `test-server.R` that the header is set, I think they are our integration tests.

I've also updated the filename to use a short id at the end so it will be `naomi_summary_<id>.zip`.

PR #74, opened 2019-11-18
mrc-726: Add logging hooks

richfitz:
This will log on all requests and might make debugging a bit easier!

Currently not tuneable (i.e., it's always on) but we can easily deal with that later.  All logs include a timestamp.


richfitz:
I don't think I've got this right for binary data...

PR #76, opened 2019-11-19
mrc-496: add generic error handler

richfitz:
This PR adds a generic error handler.  The test case I'm using is not ideal as it's really a hintr bug, but it was useful in working out how to make plumber do the right thing!


richfitz:
Logged the test work as mrc-740 and mrc-741

PR #80, opened 2019-11-21
Update tests to support including advanced run options

r-ash:
This PR will

* Update tests to reflect added advanced model run options from naomi


r-ash:
Merged into #77

PR #81, opened 2019-11-22
Add endpoint for validation of model run options

r-ash:
This PR will

* Add an endpoint to allow validation of model run options against input data

@hillalex @EmmaLRussell Added so we have the option to hook this up to the button if we have time. If not then that's fine - we can use it next year.


hillalex:
Oh right, I thought this was def not going to happen in time for this year. I have a ticket to remove the button. I don't think we'd have time anyway to do the front end work given other things are higher priority. But as you say, good to have this anyway for the future,

.travis.yml
```diff
@@ -26,7 +26,7 @@ r_packages:
 r_github_packages:
   - ropensci/jsonvalidate
   - mrc-ide/eppasm
-  - mrc-ide/naomi
+  - mrc-ide/naomi@mrc-753
```
richfitz: this can be unpinned now I believe?

docker/build
```diff
@@ -10,7 +10,7 @@ EOF
 
 rm -rf naomi
 
-git clone https://github.com/mrc-ide/naomi
+git clone --single-branch --branch mrc-753 https://github.com/mrc-ide/naomi
```
richfitz: this can be unpinned too

PR #85, opened 2019-11-27
mrc-788: Proof-of-concept translation for hintr

richfitz:
This PR adds infrastructure and examples for translating hintr responses.  Note:

* This will want changing after https://github.com/reside-ic/traduire/pull/6 is merged because the files will otherwise get very unworldly
* This does not forward the translation onto naomi (mrc-1142, then more work in hintr to do the forwarding)
* Very little is translated :grinning:


tests/testthat/test-validate-inputs.R
```diff
@@ -12,6 +12,20 @@ test_that("PJNZ can be validated and return data", {
   })
 })
 
+test_that("PJNZ can be validated and return data, in french", {
+  pjnz <- file_object(file.path("testdata", "Botswana2018.PJNZ"))
+  expect_equal(do_validate_pjnz(pjnz),
+               list(data = list(country = scalar("Botswana"),
+                                iso3 = scalar("BWA")),
+                    filters = scalar(NA)))
+  mock_read_country <- mockery::mock("GBR")
```
r-ash: This going to clash with multiple PJNZ update just done (the validation which says if GBR then fail has been removed) - I can make an equivalent test in that PR.

PR #88, opened 2019-11-28
Accept zip of PJNZ files

r-ash:
This PR will

* Add support for multiple PJNZ files
* Validate file type extensions for all files

The validation we do is
* check geojson contains a spectrum region code
* Check subnational PJNZ files are for the same country
* Check that multiple PJNZ files cannot be uploaded where more than 1 has spectrum region code of 0 - spectrum region code 0 is used to indicate a national level file so there cannot be more
* Also added a check that spectrum_region_code matches between geojson and the PJNZs but have disabled this for time being on Jeff's request and added a task for that (this won't be turned on before workshop I don't think)

Questions
* PJNZ reading is some of the slowest parts of the validation largely in how specio is working and how we're reading properties. Basically every PJNZ property read reads all the data and then pulls out the small part it is interested in and throws everything else away. Then we have to read it again the next time. Do we want to cache the PJNZ read?


.gitignore
```diff
@@ -9,3 +9,5 @@ docker/Dockerfile.worker
 inst/output/malawi_output.rds
 inst/output/malawi_spectrum_download.zip
 inst/output/malawi_summary_download.zip
+id_rsa
+tests/testthat/testdata/sensitive
```
richfitz: In order to get the tests to pass we need to do this more carefully (there might be an option [here](https://cran.r-project.org/doc/manuals/R-ints.html#Tools) to disable this check but I do not see it)

1. Add `tests/testthat/testdata/sensitive` into `.Rbuildignore` so that is not built into the package or download the data elsewhere on travis's machine
2. Add a new environment variable `NAOMI_SENSITIVE_DATA_PATH= $TRAVIS_BUILD_DIR/tests/testthat/testdata/sensitive` (goes added into the current env set, all in one ugly line)
3. Your switch then becomes something like
```r
skip_if_sensitive_data_missing <- function() {
  sensitive_path <- Sys.getenv("NAOMI_SENSITIVE_DATA_PATH", "testdata/sensitive")
  if (!file.exists(sensitive_path)) {
    skip("Sensitive data missing, check README for details.")
  }
}
```

That should enable the tests to find the data even though it's not built into the package, but still be a pleasant structure to work with locally

R/validation_asserts.R
```diff
@@ -94,6 +100,30 @@ assert_consistent_regions <- function(shape_regions, test_regions, test_source)
...
+  if (!setequal(pjnz_codes, shape_codes)) {
+    missing_shape_codes <- setdiff(shape_codes, pjnz_codes)
+    missing_pjnz_codes <- setdiff(pjnz_codes, shape_codes)
+    if (length(missing_shape_codes) > 0 &&
+        length(missing_pjnz_codes) > 0) {
+      separator <- "\n"
+    } else {
+      separator <- ""
+    }
+    stop(sprintf(
```
richfitz: This could be simplified - this message in particular, but the whole function in general

scripts/payload.json
```diff
@@ -4,8 +4,8 @@
     "shape": "tests/testthat/testdata/malawi.geojson",
     "population": "tests/testthat/testdata/population.csv",
     "survey": "tests/testthat/testdata/survey.csv",
-    "programme": "tests/testthat/testdata/programme.csv",
-    "anc": "tests/testthat/testdata/anc.csv"
+    "art_number": "tests/testthat/testdata/programme.csv",
```
richfitz: has this been reflected through the naomi metadata from https://github.com/mrc-ide/naomi/pull/30 ? that might need updating...

.gitignore
r-ash: Fixed this

R/validation_asserts.R
r-ash: Have tidied that now

scripts/payload.json
r-ash: Ran through this with naomi - the info is using the updated file names as expected

PR #90, opened 2019-11-29
mrc-1155: support for using prerun results

richfitz:
This PR create a store of prerun model results.  We use the inputs manifest in the precomputed results to identify previously run sets of data, and swap them in rather than running the model. No verification is done against the options - all valid options will count as ok here.

Associated deployment work is here: https://github.com/mrc-ide/hint-deploy/pull/25

Questions:

* should we add a comment during submission of results so that we can identify why they were added?
* should we try and swap out the mock model run to use this, and should that be done now or later?


R/prerun.R
```diff
@@ -0,0 +1,82 @@
...
+  tmp <- tempfile()
+  on.exit(unlink(tmp, recursive = TRUE))
+  zip::unzip(path, "info/inputs.csv", exdir = tmp, junkpaths = TRUE)
+  path_inputs <- file.path(tmp, "inputs.csv")
+  read_csv(path_inputs)
+}
+
+
+hash_info_inputs <- function(inputs) {
+  inputs$role[inputs$role == "art_number"] <- "programme"
```
r-ash: This isn't ideal if the rename is going to proliferate around. Not really something for this PR but I don't want to be maintaining that in multiple places. Perhaps worth thinking about if we come across having to do this renaming again.

R/run_model.R
```diff
@@ -34,15 +34,28 @@ run_model <- function(data, options, dir) {
...
+    p <- PrerunModelResults$new(path_prerun)
+    inputs <- naomi:::naomi_info_input(data)
+    if (p$exists(inputs)) {
+      message("Found prerun model results")
+      return(p$get(inputs))
+    }
+  }
+
+  path_results <- normalizePath(path_results, mustWork = TRUE)
+  dir <- normalizePath(path_results, mustWork = TRUE)
```
r-ash: This line can be removed?

r-ash:
> should we try and swap out the mock model run to use this, and should that be done now or later?

RE this I think we should, that way we only have 1 way of managing this kind of behaviour. It can probably go in a separate PR

> should we add a comment during submission of results so that we can identify why they were added?

I wonder what kind of information we would have to put here other than "model run didn't work". I think this would be useful to add but perhaps not worth adding if it is more than a small amount of effort.

R/run_model.R
richfitz: done

R/prerun.R
richfitz: As discussed on slack, this seems unavoidable at the moment

PR #91, opened 2019-11-29
Add endpoint to get barchart defaults

r-ash:
Tidy up other open PRs first before merging this

Also will need https://github.com/mrc-ide/naomi/pull/38 merged before this works


r-ash:
> Schema looks good to me. Perhaps selected_filter_options is too prescriptive, with hardcoded property names e.g. 'quarter' or 'age' might end up getting renamed. The front end just treats these as a dictionary, and doesn't care with the actual filter names are.

Yes - I meant to tidy that up. I don't know if there is a way in json schema to have an arbitrary no of properties of a fixed type? Would really like to say that it can several properties but each needs to be a FilterOption type. Will try and sort this out.

PR #101, opened 2019-12-03
mrc-1170: Add column validation

meganodris:
Added functions and tests to validate data inputs:

-  check that values in data columns meet what we expect
-  check that values in columns "year" and "calendar_quarter" are in required format
-  check that population data "source" column contains single value
-  check for when "age_group" column should be 15-49 only


r-ash:
The checks that this should do are

**Population**
* Age group contains every 5 year age group only (so not 00-00, or 15-49 etc. we see Uganda population has 00+ age groups)
* Sex is male or female
* population source is a single value
* Calendar quarter matches a regex pattern

**Survey**
* Sex is male, female or both
* Age group is a valid age group from naomi
**ANC**
* Year matches year pattern
* age group is 15-49 only
**ART**
* Sex is male, female or both
* age group is one of the naomi output age groups except 00-00 or 01-04
* Year matches year pattern

PR #106, opened 2019-12-04
Spectrum adjust

r-ash:
This PR is the equivalent to Naomi @spectrum-adjust it will

* Allow null values for mode, mean, lower and upper (Emma or Alex should say if this is okay or not)
* Update the model options
* Update delimiter for adding model options to avoid clashes with using "Age < labels"

### DIFF
--- a/R/model_options.R
+++ b/R/model_options.R
@@ -79,7 +79,7 @@ do_endpoint_model_options <- function(shape, survey, programme, anc) {
 build_json <- function(options_template, params) {
   param_env <- list2env(params, parent = .GlobalEnv)
   tryCatch(
-    glue::glue(options_template, .envir = param_env, .open = "<", .close = ">",
+    glue::glue(options_template, .envir = param_env, .open = "<+", .close = "+>",
                .transformer = json_transformer),
     error = function(e) {
       e$message <- paste0(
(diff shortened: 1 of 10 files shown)


R/model_options.R
```diff
@@ -79,7 +79,7 @@ do_endpoint_model_options <- function(shape, survey, programme, anc) {
 build_json <- function(options_template, params) {
   param_env <- list2env(params, parent = .GlobalEnv)
   tryCatch(
-    glue::glue(options_template, .envir = param_env, .open = "<", .close = ">",
+    glue::glue(options_template, .envir = param_env, .open = "<+", .close = "+>",
```
richfitz: nice - this is a good idea

tests/testthat/payload/model_submit_payload.json
```diff
@@ -21,7 +21,16 @@
...
+    "spectrum_population_calibration": "national",
+    "spectrum_plhiv_calibration_level": "subnational",
+    "spectrum_plhiv_calibration_strat": "sex_age_group",
+    "spectrum_artnum_calibration_level": "national",
+    "spectrum_artnum_calibration_strat": "age_coarse",
+    "artattend": false,
+    "rng_seed": 17,
+    "no_of_samples": 20,
+    "max_iter": 250,
+    "permissive": false
```
richfitz: does`<version_info>` below need updating? Are there any others that do too?

r-ash: I replace this as a whole thing in the test setup using `gsub` so slightly different mechanism

richfitz: ok cool - would be nice if json allowed comments as then we could leave a note to that effect :)

PR #117, opened 2019-12-09
Add 2 more new indicators

r-ash:


R/metadata.R
```diff
@@ -38,7 +38,7 @@ get_barchart_metadata <- function(output) {
     metadata$data_type == "output" & metadata$plot_type == "barchart",
     c("indicator", "value_column", "error_low_column", "error_high_column",
       "indicator_column", "indicator_value", "name")]
-  metadata[order(metadata$indicator_value), ]
+  metadata[order(as.numeric(metadata$indicator_value)), ]
```
richfitz: What are these coming as if not numeric?

r-ash: They're coming through as character from the metadata. That is because they are in a column which also has character values for some of the input indicators.

richfitz: ugh. We should fix this at some point - otherwise we're going to get NAs during conversion

PR #130, opened 2019-12-12
mrc-1246: pass language to naomi

richfitz:
This PR will pass the language to naomi, and reset appropriately.

~Will not pass until https://github.com/reside-ic/traduire/pull/11 is merged~

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -52,7 +52,7 @@ RUN install_remote \
         mrc-ide/eppasm \
         mrc-ide/heartbeatr \
         mrc-ide/rrq \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -52,7 +52,7 @@ RUN install_remote \
         mrc-ide/eppasm \
         mrc-ide/heartbeatr \
         mrc-ide/rrq \
-        reside-ic/traduire
+        reside-ic/traduire@master
```
richfitz: this was needed to force TeamCity to invalidate the build cache, but it is harmless

PR #129, opened 2019-12-12
Add missing file translation

r-ash:

PR #131, opened 2019-12-12
mrc-1249: quick and dirty translations

richfitz:
Use this instead of #129 

This PR adds rough french translations for all hintr messages

Merge after #128


docker/Dockerfile
```diff
@@ -52,7 +52,7 @@ RUN install_remote \
         mrc-ide/eppasm \
         mrc-ide/heartbeatr \
         mrc-ide/rrq \
-        reside-ic/traduire
+        reside-ic/traduire@master
```
richfitz: This is also present in #130 and was needed to force TeamCity to invalidate the build cache, but it is harmless

PR #132, opened 2019-12-13
mrc-1256: Catch and translate rrq message

richfitz:
Catch and translate string that makes it out to the front end. The note in the code is not that serious - practically the only error that will make it out is the one that is translated (and indeed that is the one in the tests). Doing better would require an overhaul of error management in rrq itself


richfitz:
Obsolete now

PR #133, opened 2019-12-16
mrc-732: Add job cancel endpoint

richfitz:
This PR adds a job cancel endpoint to hintr at `POST /model/cancel/<id>`


R/endpoints.R
```diff
@@ -94,27 +94,30 @@ endpoint_model_status <- function(queue) {
 
 endpoint_model_result <- function(queue) {
   function(req, res, id) {
-    response <- with_success(queue$result(id))
```
richfitz: this function has been refactored to convert it from a complicated set of nested if/elses to basically a switch over model status. All status except for `COMPLETE` will give an error response

R/queue.R
```diff
@@ -46,7 +46,7 @@ Queue <- R6::R6Class(
 
     status = function(id) {
       status <- unname(self$queue$task_status(id))
-      done <- c("ERROR", "ORPHAN", "COMPLETE")
+      done <- c("ERROR", "ORPHAN", "INTERRUPTED", "COMPLETE")
```
richfitz: The `INTERRUPTED` status comes from rrq

richfitz:
Agreed - I've updated those messages to be cancelled, not interrupted, on the messages that make it out from hintr

PR #139, opened 2020-01-08
Merge pull request #139 from mrc-ide/mrc-1151

r-ash:
This is in response to a bug seen by Tanzania group (amongst others) when they upload a csv file with empty rows at the bottom (i.e. the cells are empty the rows just appear as ,,,,, etc.)  then they get validation errors like

image

We strip empty rows before using the data - we need to check what the implication for naomi is here and fix there too if necessary


R/model_options.R
```diff
@@ -83,7 +83,7 @@ build_json <- function(options_template, params) {
     glue::glue(options_template, .envir = param_env, .open = "<+", .close = "+>",
                .transformer = json_transformer),
     error = function(e) {
-      e$message <- t_("MODEL_OPTIONS_FAIL", list(message = message))
+      e$message <- t_("MODEL_OPTIONS_FAIL", list(message = e$message))
```
richfitz: This change implies that the previous behaviour was broken - could you add a quick test for it now it's fixed please?

R/read_data.R
```diff
@@ -20,7 +20,9 @@ read_csv_regions <- function(csv_file) {
 }
 
 read_csv <- function(...) {
-  utils::read.csv(..., stringsAsFactors = FALSE)
```
richfitz: Do we want to just strip *trailing* empty rows?  This removes internal empty rows too, which might be ok, but is more invasive.

tests/testthat/test-read-data.R
```diff
@@ -57,3 +57,8 @@ test_that("can read spectrum region code from geojson", {
   region_codes <- read_geojson_spectrum_region_codes(shape)
   expect_equal(region_codes, c(10, 11, 12, 13, 14, 15, 16, 17, 18, 19))
 })
+
+test_that("read csv removes empty rows", {
+  data <- read_csv(file.path("testdata", "population_empty_rows.csv"))
+  expect_equal(nrow(data), 2448)
```
richfitz: Perhaps add a smaller test file :)

You could just write one though (untested):

```
path <- tempfile()
writeLines(c("a,b", "1,2", "3,4", "5,6", ",", ","), path)
expect_equal(nrow(read_csv(path)), 3)
```

R/read_data.R
r-ash: Yep intending to move any empty rows. I haven't seen any examples with empty rows in the middle of the data but I think this would not have meaning. Jeff is stripping these rows in naomi so I think this is the correct approach.

tests/testthat/test-read-data.R
r-ash: Way better - thanks.

R/model_options.R
r-ash: Yep - I've expanded a test we previously had written for this (it was only checking the first half of it though then matching on \w+) I've updated that so it checks it matches the entire string.

PR #144, opened 2020-01-14
Speed up travis build

r-ash:
This PR will
* Split code coverage and tests into separate build stages run simultaneously

We talked a bit about how we could speed up the builds we have roughly
* 5 min installing (if using R package cache)
* 7 min running R CMD check
* 7 min running coverage

Splitting these should mean if we have caching that the development cycle (install + test) should give us feedback in 12 min over 19 which is a bit better.

Other stuff we could look at doing
* Caching failed builds still - if there is no R package cache and the build fails the next time it builds it will go through installing the packages from source again. We should cache these on failed builds too. (is that right?)
* Split unit & integration tests - R has no idea of an integration test but we kind of have one here in `test-server.R` script. We could look at splitting this from the rest of the tests as it is quite a slow one. I'm not sure on how big a gain this would be as I think I would be hesitant to deploy to production without seeing this test pass
* Save out build R package for testing and coverage see https://docs.travis-ci.com/user/build-stages/#data-persistence-between-stages-and-jobs (would take us down to a 5 min build then 2 simultaneous 7 min builds - not actually any quicker overall)
* Speeding up naomi build too. The current cycle for getting a naomi build deployed is:
   * naomi change gets pushed which triggers naomi tests (~ 20 min to run)
   * Make a corresponding hintr PR to update version number which triggers hintr tests (~ 13min with this PR)
   * Wait for both builds to pass (if they don't - fix and do the whole thing again) 
   * Deploy to staging and run through app
   * Merge naomi PR
   * Update pins on hintr PR (to use master branch of naomi again) and wait for builds to pass (~ 13 min for builds but sometimes we're skipping waiting for this to be done)
   * Merge hintr PR into master and wait for master teamcity build (5 min with cache or 30 min without)
   * Redeploy production

So really at a minimum we have a cycle time of 20 + 5 + checking time. Probably around 30 mins total if we have packages cached and we don't wait for the second hintr builds to pass.
Perhaps next best thing to do is to split the naomi build into test + covr like we have here?


r-ash:
Closing this as we've moved to buildkite now

PR #159, opened 2020-01-22
mrc-1301: debug endpoint for hintr

richfitz:
This PR will create a "debug" endpoint, with which we can download all the inputs for running a model. The stack trace will include the model run id and that is the information that we need to get from the participants.

For deployment, requires https://github.com/mrc-ide/hint-deploy/pull/28 for exposing the hintr port to the server - we can connect to that via the api.

I will add client support to naomi


R/endpoints.R
```diff
@@ -108,7 +108,8 @@ endpoint_model_result <- function(queue) {
                        value = process_result(queue$result(id)))
     } else if (task_status == "ERROR") {
       result <- queue$result(id)
-      error_data <- structure(result$message, trace = result$trace)
+      trace <- c(sprintf("# %s", id), result$trace)
```
richfitz: This line is important - it adds the model run id to the trace so that we can get it out of the user session without dealing with devtools

R/endpoints.R
```diff
@@ -122,6 +123,36 @@ endpoint_model_result <- function(queue) {
...
+
+    data <- response$value
+    files <- vcapply(data$objects$data, identity, USE.NAMES = FALSE)
+    tmp <- tempfile()
+    path <- file.path(tmp, id)
+    dir.create(path, FALSE, TRUE)
+
+    path_files <- file.path(path, "files")
+    dir.create(path_files)
+    file_copy(files, file.path(path_files, basename(files)))
```
richfitz: We copy the files in as their basename only. In case of duplicate files that's going to be a bit ugly. If all 6 files are not required, then this will fail - can you comment @r-ash?

tests/testthat/helper-server.R
```diff
@@ -86,3 +86,12 @@ hintr_server <- function(n_tries = 10, poll = 0.5, results_dir = tempdir()) {
 response_from_json <- function(x) {
   jsonlite::parse_json(httr::content(x, "text", encoding = "UTF-8"))
 }
+
+hintr_server_blocking <- function() {
```
richfitz: I was using this for debugging a server error - it just bring up hintr so that I could use browser within the server session

tests/testthat/test-server.R
```diff
@@ -623,6 +650,7 @@ test_that("404 pages have sensible schema", {
 })
 
 test_that("Error handler is triggered", {
+  testthat::skip("This is fixed now, but this test should be rewritten")
```
richfitz: This test no longer works because it relied on buggy behaviour in the zip serializer which is now fixed.

R/endpoints.R
```diff
@@ -122,6 +123,36 @@ endpoint_model_result <- function(queue) {
...
 }
 
+endpoint_model_debug <- function(queue) {
+  function(req, res, id) {
+    response <- with_success(queue$queue$task_data(id))
+    if (!response$success) {
+      response$errors <- hintr_errors(list(
+        "INVALID_TASK" = response$message))
+      res$status <- 400
+      return(hintr_response(response, "ModelDebugResponse"))
```
r-ash: Is this schema in this PR? I'm surprised this doesn't fail some tests.

R/endpoints.R
```diff
@@ -122,6 +123,36 @@ endpoint_model_result <- function(queue) {
...
+
+    path_files <- file.path(path, "files")
+    dir.create(path_files)
+    file_copy(files, file.path(path_files, basename(files)))
+    saveRDS(data, file.path(path, "data.rds"))
+
+    on.exit(unlink(tmp, recursive = TRUE))
+
+    dest <- paste0(id, ".zip")
+    dest_full <- file.path(tmp, dest)
```
r-ash: `dest_full` is not used here

R/endpoints.R
r-ash: Duplicate files might not necessarily fail validation. Thinking that they could upload the same file for the programme or survey data (as long as all the expected rows from every file are available in what they upload). We don't expect to ever get a duplicate file but they could potentially do it. It would take some effort though.

I don't think we will ever get an occasion where all 6 files are not required. Doing a model run with no data would be pretty meaningless? At least as we're currently working 4 files are required and I'm not expecting that to change.

R/debug.R
```diff
@@ -0,0 +1,35 @@
...
+##' @param server The url of the server.  The default is to use the
+##'   production naomi/hint/hintr instance. It is not possible to use
+##'   the staging instance.  You can change this if running locally.
+##'
+##' @param dest The destination for the downloaded data.  The actual
+##'   data will be unpacked into a directory corresponding to the run
+##'   id within this, so it is safe to use a common directory.
+##'
+##' @export
+download_debug <- function(id, server = NULL, dest = tempfile()) {
```
r-ash: This will mean that anyone with the model run ID will be able to retrieve the input data from the workshops? Is the security that they have know the model run ID enough?

tests/testthat/test-server.R
r-ash: Should we just remove it?

R/endpoints.R
richfitz: Oh yeah, this just does not get triggered I think because of the error handler. I was a bit confused about other cases here (and we don't really return json)

R/debug.R
richfitz: yes - chance of guessing the ID is close to zero, though we are passing it around over http rather than https

PR #161, opened 2020-01-22
mrc-1310: Add a progress bar

richfitz:
This PR adds a progress bar to the debug download output, which is considerably less cryptic when it takes a while, which it typically will.

Jeff - you can test by running:

```
p <- hintr::download_debug(id)
```

with the id I send you over slack, while on the VPN, and you can download an error that Megan saw earlier.  If you are not on the VPN then you'll get a curl timeout error after 10s.  The value `p` will be the path to the downloaded files, the formats of which will be fairly self explanatory (that functionality was all merged in #159)

### DIFF
--- a/R/debug.R
+++ b/R/debug.R
@@ -13,8 +13,11 @@
 ##'   data will be unpacked into a directory corresponding to the run
 ##'   id within this, so it is safe to use a common directory.
 ##'
+##' @param verbose Add a progress bar
+##'
 ##' @export
-download_debug <- function(id, server = NULL, dest = tempfile()) {
+download_debug <- function(id, server = NULL, dest = tempfile(),
+                           verbose = TRUE) {
   if (is.null(server)) {
     server <- "http://naomi.dide.ic.ac.uk:8888"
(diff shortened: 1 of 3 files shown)


richfitz:
For your points:

1. This will be easy to do do and I've made a ticket here: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1312 - I'll do that when we get a few minutes of downtime
2. I can output the sessionInfo() at the same time, though practically it's not that often changing on the server and is easily fetched.  The downside of adding it is somewhat larger downloads, but this seems small - I've left a comment to do that with mrc-1312
3. This one requires us to keep better tabs on the filenames and while I agree that it would be better it requires more work in hintr than is going to happen in the short term.  Longer term this would be nice, and I think there is an issue somewhere for other places where this information would be useful (the download endpoint perhaps?)

jeffeaton:
Sounds good -- thanks. On (3) Rob addressed it for the download so that the file list in `/info` does return the actual file name now, so that need is sorted.

PR #179, opened 2020-04-06
mrc-1471: Add buildkite support

richfitz:
This PR adds buildkite support following the same pattern as the orderly chain, with one difference - in `docker/build` we respond to the environment variable `$NAOMI_SHA` to check out the appropriate version of naomi to build against.  Apart from that this alters the scripts minimally to cope with switching agents, and to be robust to directory changes (we have a mix of assuming we're running from the root directory and being robust to that, which is weird).

We can remove the support for teamcity entirely from this build, or in a separate PR.  I think that this (plus the companion PR for hintr) also remove the need for the naomi bot

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -11,3 +11,4 @@
 ^id_rsa\.enc$
 ^id_rsa$
 ^tests/testthat/testdata/sensitive$
+^buildkite$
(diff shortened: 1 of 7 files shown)


richfitz:
Oh good point - let's think about that in a future bit of work - can you think about this at some point?

PR #183, opened 2020-06-22
Various changes to support pkgapi re-implementation

r-ash:
Added this PR to tidy up some of the impl we need for hintr2. Originally to update impl of `scalar` so that it won't doubly apply the type but this grew as more problematic things reared their head. 

* `scalar` won't reapply scalar class if it already exists on the value
* Also fixes some other bugs noticed (I think introduced because of package updates)
* update functions which return `scalar(NA)` to instead use `scalar(json_verbatim(NULL))` see https://github.com/mrc-ide/hintr2/pull/1#discussion_r431974990 for a discussion
* Update to naomi v 1.0.0 (which removes INLA and tmbstan from naomi) see https://github.com/mrc-ide/naomi/pull/119


scripts/run_model
```diff
@@ -21,10 +21,6 @@ invisible(lapply(payload$data, function(x) {
   }
 }))
 
-## Don't use dynamic libaries as these don't currently work on travis with
-## up to date INLA build
-INLA:::inla.dynload.workaround()
```
richfitz, line 26: 🎉

tests/testthat/test-model-options.R
```diff
@@ -2,18 +2,18 @@ context("model-options")
 
 test_that("can build JSON from template", {
   json <- build_json("test <+param1+>", list(param1 = scalar("test string")))
-  expect_equal(json, 'test "test string"')
+  expect_equivalent(json, 'test "test string"')
```
richfitz, line 5: What attributes mismatch here now?

r-ash: Glue has changed and now outputs from `glue::glue` have class `glue`

richfitz: that's so weird. Why would you want that? `sprintf()` does not return objects with class "sprintf"...

PR #191, opened 2020-09-16
Add test and fix for file naming bug

r-ash:
Reported a bug where a user downloading spectrum digest from an old model run was getting a filename like

`_20200916-103008_naomi_spectrum_digest.zip`

This is because we try and prefix with the country the run is for, this info comes from metadata in the response from the model run and hasn't always been available. So in old runs where this metadata isn't there then it produces a string as above with this blank. Really what should happen is there is no preceding `_` if that country info isn't available.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hintr
 Title: R API for calling naomi district level HIV model
-Version: 0.1.0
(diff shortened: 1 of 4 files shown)


jeffeaton (approved):
Looks good -- thanks for tidying this.

PR #193, opened 2020-09-22
mrc-1744 expect and return fromADR flag

hillalex:
The `SessionFile` objects now contain an extra property `fromADR` - at least for now all hintr has to do is pass this flag back out with the validation response.


hillalex:
Actually sorry hold off on merging for now until HINT supports this!

hillalex:
In adding the extra "fromADR" flag to the input object I discovered a bug with the validation messages.

We were getting back messages like:
"Le fichier doit être de type PJNZ, zip, mais il est du type geojson,geojson,geojson."

Which on addition of the ADR flag became 
"Le fichier doit être de type PJNZ, zip, mais il est du type geojson,geojson,geojson,."

Instead of the expected:
"Le fichier doit être de type PJNZ, zip, mais il est du type geojson."

The reason is that the `assert_file_extension` function was written to assume a scalar argument, but actually being passed a list. The tests didn't pick this up because `expect_error` matches a regex, so 

`expect_error(..."File must be of type PJNZ, zip, got type csv.")`

Was passing even though the error message was actually: 
"File must be of type PJNZ, zip, got type csv,,,."

This commits fixes the tests and the error messages:

https://github.com/mrc-ide/hintr/pull/193/commits/b252f891b0f68653e8aefae48bbac50ef7ef3471

hillalex:
Ok, this is ready for review now!

r-ash (approved):
Thanks for fixing the bug!

PR #194, opened 2020-09-22
Fix test failure and pin to updated base image

r-ash:
Some libs have changed which is causing problems with the builds

Fixing here and in the base image naomi.docker, will remove pinned docker base branch once this is working


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mrcide/naomi-base:latest
+FROM mrcide/naomi-base:add-latex-lib
```
r-ash, line 1: I need to unpin this after https://github.com/mrc-ide/naomi.docker/pull/8 has been merged

tests/testthat/helper-server.R
```diff
@@ -62,6 +62,7 @@ hintr_server <- function(n_tries = 10, poll = 0.5, results_dir = tempdir()) {
       TRUE
     }, error = function(e) FALSE)
     if (ok) {
+      withr::defer_parent(process$kill())
```
r-ash, line 65: Manually killing this instead of relying on garbage collect I hope will help with memory problems

tests/testthat/test-endpoints-model.R
```diff
@@ -281,12 +281,13 @@ test_that("model run can be cancelled", {
   expect_true("id" %in% names(response))
   id <- response$id
 
-  running <- queue$queue$worker_task_id()
-  expect_equal(scalar(unname(running)), id)
```
r-ash, line 285: this was flaky and not testing anything useful so I removed it

richfitz (approved):

PR #195, opened 2020-09-28
MRC-1790: Update name from download summary to download coarse outputs

r-ash:
This is in preparation for adding an endpoint for download summary report to avoid ambiguity between existing endpoint

This is a breaking change for front end

Also the existing endpoint naming is not really representative of the actual download. If I remember we added summary endpoint expecting a report impl on science side but this never got finished in time but they hacked in another output which people wanted. 

We shouldn't merge this until the front end has a PR which builds off this one and works with the updated name. We also shouldn't merge until https://github.com/mrc-ide/naomi/pull/124 has been merged

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -12,3 +12,4 @@
 ^id_rsa$
 ^tests/testthat/testdata/sensitive$
 ^buildkite$
+^naomi$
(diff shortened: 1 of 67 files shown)


EmmaLRussell (commented):
Thanks Rob, I'll make a start on putting that change into HINT

hillalex:
https://github.com/mrc-ide/hint/pull/346 now merged

richfitz (approved):

PR #200, opened 2020-10-01
Fix memory problems

r-ash:
Special 200th PR :confetti_ball: :champagne: :partying_face: 

This PR will
* Make sure workers are killed
* Workers spawned by queue have a timeout of 5 mins

This (hopefully) fixes all the annoying memory issues I've had working with hintr. The problem was in particular tests where they start a job on a worker and then the test completes before the job finishes the call to `cleanup` would not kill the worker processes. This meant after running test suite you would end up with several orphaned worker processes using up memory


richfitz (approved):
:tada:

PR #203, opened 2020-10-20
Naomi dev 2021

jeffeaton:
This branch are updates to hintr to align with the 2021 version of Naomi (naomi >=2.0).

I'm opening a PR for this as a place for discussion, but not merging until we're ready to deploy the new version.  (Please suggest if you suggest a different approach for this.)


tests/testthat/test-filters.R
```diff
@@ -191,8 +191,7 @@ test_that("error thrown when tree can't be constructed", {
 })
 
 test_that("naomi IDs can be mapped to hint IDs", {
-  expect_equal(get_hint_id(2), "prevalence")
-  expect_equal(get_hint_id("2"), "prevalence")
+  expect_equal(get_hint_id("prevalence"), "prevalence")
```
jeffeaton, line 194: @r-ash -- this change is because the indicators metadata table is updated with `indicator` rather than `indicator_id` ([here](https://github.com/mrc-ide/naomi/commit/69e880561832a8fbf42d4c9ab65dd7e00b637b99)). Can you review this change and make sure it is the right thing to do and flag if any implications for hint?

jeffeaton:
Note that 7354d3b updates a few of the metadata schemas:

* `inst/schema/BarchartIndicator.schema.json`
* `inst/schema/ChoroplethIndicatorMetadata.schema.json`
* `inst/schema/ModelResultRow.schema.json`

Make sure to flag these to HINT when merging.

tests/testthat/test-filters.R
r-ash: This function and `get_indicator_display_name` aren't actually called from anywhere - must be old helpers which never got removed. I've removed them now

r-ash (approved):
This looks fine to me, I think should work fine with hint as the changes are mostly in data structure which is metadata driven. I'll add someone on the hint side as reviewer though as they generate interfaces from the json schema so might be some updates needed there.

PR #204, opened 2020-10-21
update pkgapi dependency to >= 0.0.10

jeffeaton:
pkgapi_v0.0.10 updated some things for plumber 1.0.0, which helped clear some errors I was getting on the tests.


r-ash (approved):
Thanks for this

PR #208, opened 2020-11-11
update ART data schema column year to calendar_quarter

jeffeaton:
We agreed last week to update the ART programme data schema to field `calendar_quarter` instead of `year`.  I thought that `hintr` was already equipped for this, but it is not. It is requiring column `year` right now.

This PR updates the schema to column `calendar_quarter`.  I'm having trouble running the tests locally again, and suspect this may fail some tests.

This is blocking model testing, so it would be great if we can try to work through and deploy quickly.

Thanks,
Jeff


jeffeaton:
We've got a problem here (which @r-ash mentioned last week...).

The function `do_validate_programme()` returns `"year"` filters for HINT to population the chloropleth: https://github.com/mrc-ide/hintr/blob/428777a9b824c90f6e81e69e9a84b9b1bb356750/R/validate_inputs.R#L131.

We've now switched to the `calendar_quarter` field in this dataset.  This needs to be replaced with `get_quarter_filters(data)`.  I'm not sure what the implications of this for HINT are though.

Is it possible to (1) return the quarter labels here instead of year labels, and (2) change the label on the user interface to "Period" instead of "Year"?

Thanks,
Jeff

jeffeaton:
I think this should be ready now. I updated the line mentioned above to return the quarter labels to the year field:

https://github.com/mrc-ide/hintr/blob/0a07082e163851982cf83965b73cc4909433794d/R/validate_inputs.R#L131

And updated tests accordingly.

r-ash (approved):

PR #212, opened 2020-11-26
Add model options endpoint for calendar_quarter_t1_default

jeffeaton:
This adds an endpoint for `calendar_quarter_t1_default` to return the most recent `survey_mid_calendar_quarter` from the surveys dataset as a default for the `calendar_quarter_t1` option.

It corresponds to PR mrc-ide/naomi#162.

😨


R/model_options.R
```diff
@@ -52,6 +54,7 @@ do_endpoint_model_options <- function(shape, survey, programme, anc) {
     area_level_options = area_level_options,
     area_level_default = area_level_options[[length(area_level_options)]]$id,
     calendar_quarter_t1_options = time_options,
+    calendar_quarter_t1_default = most_recent_survey_quarter,
```
r-ash, line 57: I think we need to make sure that `most_recent_survey_quarter` is in the list of `time_options` it probably always is? Time options is a fixed list of quarters from `Jan 2010` to now. I think putting a line to add `most_recent_survey_quarter` to the `time_options` if it isn't already in there would be worth while

r-ash (changes requested):
Changes look good, there are some test failures https://buildkite.com/mrc-ide/hintr/builds/423#2ea9299d-7254-4a6f-86ae-564edb5b520a/30-165 these might just be because of issues in naomi json or the version of naomi being out of date.
Can you also
* Bump the version of this package
* Update dependency on naomi verison to `2.0.6` 
* Update the docker build to build the open branch of naomi with this change in see https://github.com/mrc-ide/hintr/commit/ecca118df23313ef76db1300c499c52edd9bb43c for an example of what this looks like

I can then deploy to staging when the build passes

R/model_options.R
jeffeaton: Thanks good shout. Addressed in 1505cc6.

r-ash (approved):

PR #215, opened 2020-12-03
Turn off multithreading for TMB

r-ash:
Not sure how to test this, running locally without this change it isn't running on more than 1 CPU and staging has only 1 CPU available as it is a vagrant box. Open to suggestions. I can at least test on staging to make sure it doesn't break anything.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,5 +32,8 @@ RUN R CMD INSTALL /src
 
 EXPOSE 8888
 ENV HINTR_QUEUE_ID=hintr
+## Model run will try to parallelise over as many threads as are available
+## potentially slowing the application, manually limit threads to 1
+ENV OMP_NUM_THREADS=1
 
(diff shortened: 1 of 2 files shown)


r-ash:
Have deployed this on staging and confirmed that the model still runs so we haven't broken entrypoint for the workers. I can see the call to `TMB::openpm` return value in the logs too so I have confidence this is being run.

richfitz (commented):
Once you confirm that TMB used to spike the CPU without this (on production or elsewhere) and that this change fixes it, we can merge.

r-ash:
Turns out using `TMB::openmp(1)` didn't work and we saw on production this still using 10 cores. After more digging, using OMP env var `OMP_NUM_THREADS` does work to limit this. I have checked this by running a model manually on production, both with and without the flag and confirmed it works.

richfitz (approved):

PR #216, opened 2020-12-04
Validate survey data survey_mid_calendar_quarter column

jeffeaton:
This PR:
* Revises the functions `assert_calendar_quarter_column()` and `assert_year_column()` to accept an argument `col_type = ` to apply the validation to columns with a different name but same specification.
* Adds assertion check to for `survey_mid_calendar_quarter` column in `do_validate_survey()`

This would catch the error from Tuesday just before the UNAIDS ToT due to misspecified survey_mid_calendar_quarter upon data upload.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+# hintr 0.1.14
+
+* Generalize functions `assert_calendar_quarter_column()` and `assert_year_column()` to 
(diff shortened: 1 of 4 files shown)


r-ash (approved):
Thanks, this looks good

PR #217, opened 2020-12-04
Naomi 2.1.1

jeffeaton:
This is pinned to naomi@option-output-aware (PR 178 on Naomi), which:

* adds an advanced option "output_aware_plhiv" and
* makes advanced options block collapsible


jeffeaton:
This is a regrettable mess. Closing it but not deleting it so we can remember how bad things can get.

PR #218, opened 2020-12-04
pin to naomi@option-output-aware

jeffeaton:
Let's try this again. This is pinned to naomi@option-output-aware (mrc-ide/naomi#178).

* adds an advanced option "output_aware_plhiv" and
* makes advanced options block collapsible

The ModelRunOptions.schema needs `collapsible` and `collapsed` added: 
https://github.com/mrc-ide/hintr/blob/master/inst/schema/ModelRunOptions.schema.json#L73

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hintr
 Title: R API for calling naomi district level HIV model
-Version: 0.1.13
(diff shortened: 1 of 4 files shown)


r-ash:
see #219 (this was failing because branch name is `origin/...` and docker doesn't like that

PR #219, opened 2020-12-04
Naomi option output aware

r-ash:
see #218


jeffeaton (approved):
Looks great, if i do say so mysefl.

PR #222, opened 2020-12-05
Naomi awareness of HIV status indicators (v2.1.2)

jeffeaton:
This linked to mrc-ide/naomi#184.

It pins to pins to branch naomi@aware-of-status and updates the hintr tests for new model options in Naomi and additional output indicators.


tests/testthat/test-api.R
```diff
@@ -1,1217 +1,1217 @@
-context("api")
-
-test_that("don't change language if not asked to", {
```
r-ash, line 3: :face_with_head_bandage:

typescript
```diff
@@ -0,0 +1 @@
+Script started on Wed Dec  2 20:35:01 2020
```
r-ash, line 1: Not sure where this is from?

r-ash (changes requested):

typescript
jeffeaton: 🤷

jeffeaton:
936b88a adds new options to the validate options JSON payload, which should fix those erroring API tests.

I noticed a comment about model_options_valdate() 'endpoint not currently called'. I think that is out of date now?

https://github.com/mrc-ide/hintr/blob/39883b573abeecf3e3816fb7ca6c7e923b6ec839/R/endpoints.R#L107-L112

r-ash (approved):

PR #223, opened 2020-12-07
Update naomi version no and pin to branch

r-ash:


jeffeaton (approved):
Not much to complain about here.  Do you want to bump the hintr version number, or no need if it's only a branch pin?

PR #225, opened 2020-12-08
Populate new default model options from Naomi

r-ash:
This felt a bit harder than I think it should have been. I think entropy taking its toll on this part of the app, we should refactor when we have more time.


tests/testthat/test-model-options.R
```diff
@@ -64,12 +64,17 @@ test_that("do_endpoint_model_options correctly builds params list", {
                  "calendar_quarter_t1_options",
                  "calendar_quarter_t1_default",
                  "calendar_quarter_t2_options",
-                 "survey_prevalence_options", "survey_art_coverage_options",
+                 "survey_prevalence_options", "survey_prevalence_default",
+                 "survey_art_coverage_options", "survey_art_coverage_default",
```
jeffeaton, line 68: Putting these on same line is discrepant with similar cases before and after. Suggest one each line.

tests/testthat/test-model-options.R
```diff
@@ -147,12 +159,17 @@ test_that("do_endpoint_model_options without programme data", {
...
```
jeffeaton, line 163: separate lines

tests/testthat/test-model-options.R
```diff
@@ -411,14 +449,69 @@ test_that("model options work when survey_mid_calendar_quarter missing", {
...
```
jeffeaton, line 453: separate lines

jeffeaton (approved):
Looks good. Only comments are code styling in the tests -- take or leave.

PR #230, opened 2020-12-10
Update test data & tests

r-ash:
This replaces the test data we use to build the mock model with the data from the ADR. We need these to be the same dataset because the front end uses the mock model in their local testing & they need it to align with the uploaded data (which comes from ADR) because otherwise metadata is out of sync with model response data.

Adding @jeffeaton  as a reviewer too for if this data changes again


EmmaLRussell (approved):
Works for me! :+1:

PR #234, opened 2020-12-11
Notify teams

r-ash:
This is a better notification system for us than doing it through me forwarding emails from my account

Couple of things here
* This notifys a webhook which a microsoft flow is watching, the flow then processes output into a card and sends it to teams
* Do you want to review the teams card whilst we are here? This is pretty flexible and can add content that we want
* I don't know if the conditional on notify works here - we can alternatively manage this conditional behaviour in the microsoft flow thing 
* I would like to have it notify failed & fixed builds ideally but not really sure what the `build.state` from buildkite is for these

TODO
* Secure the address here instead of having it public
* Add notes to knowledge base
* See if notify failed & fixed builds is working
* Happy face for fixed builds


buildkite/pipeline.yml
```diff
@@ -17,3 +17,7 @@ steps:
 
   - label: ":shipit: Push images"
     command: docker/push
+
+notify:
```
richfitz, line 21: can you investigate if these are safe to leave in public code? if not I think that buildkite does environment variable interpolation, so we could redact the token part of this (between `workflows` and `triggers`) and replace with an token that we provision into the machine - and then revoke this one

richfitz (changes requested):

r-ash:
Closing this as instead have moved to sending notification using settings in buildkite. See knowledge base article in youtrack for details.

PR #237, opened 2020-12-14
MRC-2050: Implement calibrate submit endpoint

r-ash:
This builds on #233  and should be merged into mrc-2040 after that PR


tests/testthat/test-endpoints-model.R
```diff
@@ -357,168 +357,3 @@ test_that("Debug endpoint errors on nonexistant id", {
                scalar("Task '1234' not found"))
   expect_equal(error$status_code, 400)
 })
-
-test_that("can calibrate a model result", {
```
r-ash, line 361: These have moved to `test-calibrate` and been updated

R/endpoints.R
```diff
@@ -158,6 +158,23 @@ model_result <- function(queue) {
...
+    verify_result_available(queue, id)
+    calibration_options <- jsonlite::fromJSON(input)
+    if (!is_current_version(calibration_options$version)) {
+      hintr_error(t_("CALIBRATE_SUBMIT_OLD"), "VERSION_OUT_OF_DATE")
+    }
+    tryCatch(
+      list(id = scalar(queue$submit_calibrate(queue$result(id),
+                                              calibration_options$options))),
+      error = function(e) {
+        hintr_error(e$message, "FAILED_TO_QUEUE")
```
richfitz, line 172: when does this happen?

richfitz (commented):
This looks super tidy - just one comment about failure to queue. I can't immediately see where that would be thrown but as it's not showing up as a coverage gap it must be?

R/endpoints.R
richfitz: Ah, re: coverage - we don't have coverage enabled here. We have a workflow for tokens now

richfitz (approved):

PR #246, opened 2021-01-05
MRC-2077: Queue calibrate tasks as higher priority than model run

r-ash:
Depends on https://github.com/mrc-ide/rrq/pull/27 being merged first

### DIFF
--- a/R/main.R
+++ b/R/main.R
@@ -40,7 +40,8 @@ hintr_worker [<queue_id>]"
 main_worker <- function(args = commandArgs(TRUE)) {
   # nocov start
   rrq::rrq_worker(hintr_queue_id(main_worker_args(args)$queue_id, TRUE),
-                  heartbeat_period = 10)
+                  heartbeat_period = 10,
+                  queue = c("calibrate", "run"))
   # nocov end
 }
 
--- a/R/main.R
+++ b/R/main.R
@@ -41,7 +41,7 @@ main_worker <- function(args = commandArgs(TRUE)) {
   # nocov start
   rrq::rrq_worker(hintr_queue_id(main_worker_args(args)$queue_id, TRUE),
                   heartbeat_period = 10,
-                  queue = c("calibrate", "run"))
+                  queue = c(QUEUE_CALIBRATE, QUEUE_RUN))
   # nocov end
 }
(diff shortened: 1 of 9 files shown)


R/main.R
```diff
@@ -40,7 +40,8 @@ hintr_worker [<queue_id>]"
 main_worker <- function(args = commandArgs(TRUE)) {
   # nocov start
   rrq::rrq_worker(hintr_queue_id(main_worker_args(args)$queue_id, TRUE),
-                  heartbeat_period = 10)
+                  heartbeat_period = 10,
```
richfitz, line 43: We might look at another argument for a worker that *only* does calibrations? That would require changes to the deploy too, and I think worth thinking about in separate tickets/PRs

R/queue.R
```diff
@@ -22,7 +22,8 @@ Queue <- R6::R6Class(
       message(t_("QUEUE_STARTING"))
       queue_id <- hintr_queue_id(queue_id)
       self$queue <- rrq::rrq_controller(queue_id, con)
-      self$queue$worker_config_save("localhost", heartbeat_period = 3)
+      self$queue$worker_config_save("localhost", heartbeat_period = 3,
+                                    queue = c("calibrate", "run"))
```
richfitz, line 26: Do we have a string constant set anywhere? Replacing these string literals with `queue_calibrate` and `queue_run` (and `queues <- c(queue_calibrate, queue_run)`) might help with some typos

You can also use the default queue for run, which might be simpler?

richfitz (approved):
Oh wow this is gratifyingly small and clean. Did you try bringing up the whole lot and seeing if it works in practice as this seems too good to be true :)

richfitz (changes requested):
sorry, butterfingers

R/main.R
r-ash: Added a ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2099

r-ash:
Looks like integration test is flaky in most recent run which is a bit odd.

r-ash:
Have confirmed that I can see this working locally

richfitz (approved):

PR #252, opened 2021-01-22
mrc-2133: Run model in a separate process using callr

r-ash:
Builds on #254 

This will use rrq to run model run in a separate process hopefully in an attempt to reduce memory leak we've seen on production.

If a model run errors during running then it leaves memory occupied which doesn't get cleaned up, leading to bad failure as we have seen on production. Using callr when the process exits it should clean up memory.

Note this doesn't run calibration in a separate process as an attempt to avoid the overhead of setting up and reading results out. Have noticed though that it looks like the memory is creeping up when I do this, after a calibration run the docker container is using ~200MB more than previous and manually running gc on the worker using `message_send_and_wait` isn't clearing this. That probably means another leak right?


r-ash:
> Note this doesn't run calibration in a separate process as an attempt to avoid the overhead of setting up and reading results out. Have noticed though that it looks like the memory is creeping up when I do this, after a calibration run the docker container is using ~200MB more than previous and manually running gc on the worker using `message_send_and_wait` isn't clearing this. That probably means another leak right?

Update to above comment, calibration now runs in a separate process too. Overhead to do that feels small compared to length of running the task but admittedly I haven't timed them. But this has cleared up the memory behave I mentioned above

richfitz (approved):

PR #254, opened 2021-01-22
Save calibrate outputs to new files

r-ash:
Corresponding naomi PR https://github.com/mrc-ide/naomi/pull/213

### DIFF
--- a/R/calibrate.R
+++ b/R/calibrate.R
@@ -1,4 +1,5 @@
-run_calibrate <- function(model_output, calibration_options, language = NULL) {
+run_calibrate <- function(model_output, calibration_options, path_results,
+                          language = NULL) {
   if (!is.null(language)) {
     reset_hintr <- traduire::translator_set_language(language)
     reset_naomi <-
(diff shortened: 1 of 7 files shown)


R/calibrate.R
```diff
@@ -9,5 +10,14 @@ run_calibrate <- function(model_output, calibration_options, language = NULL) {
     })
   }
 
-  naomi::hintr_calibrate(model_output, calibration_options)
+  path_results <- normalizePath(path_results, mustWork = TRUE)
+  output_path <- tempfile(tmpdir = path_results, fileext = ".rds")
```
richfitz, line 14: To save yourself a headache here I'd add a prefix.

```
tempfile("output", path_results, ".rds")
```

which and similarly for the other output types

richfitz (changes requested):

R/calibrate.R
r-ash: Thanks, have prefixed those here an in model outputs

richfitz (approved):

PR #273, opened 2021-04-27
Update references to use new rrq

richfitz:


tests/testthat/test-endpoints-model.R
```diff
@@ -218,15 +218,15 @@ test_that("model run can be cancelled", {
...
+    expect_true("CANCEL" %in% log$command)
+    expect_equal(queue$queue$task_status(id), setNames("CANCELLED", id))
   })
 
   get_status <- queue_status(queue)
   response <- get_status(id)
   expect_true(response$done)
   expect_false(response$success)
-  expect_equal(response$status, scalar("INTERRUPTED"))
+  expect_equal(response$status, scalar("CANCELLED")) # TODO: update schema?
```
r-ash, line 229: Old TODO here, looks like status is just a `string` in schema so I don't think anything to change.

r-ash, line 229:
```suggestion
  expect_equal(response$status, scalar("CANCELLED"))
```

r-ash (changes requested):
One suggested change but this looks quite straight forward

r-ash (approved):

PR #274, opened 2021-05-07
Add dummy endpoint to return plotting data for calibration

r-ash:
This PR will
* Add a new endpoint `calibrate/plot/<id>` which returns dummy data and metadata of format needed for plotting calibrated data comparison bar chart

I think one outstanding question on structure, at the moment the x axis is fixed as is what we disaggregate by. But in the standard bar chart we are relying on returning filters for the mapping from data to label for all things. So if return region as a filter that is confusing because we don't actually want to filter on it, but if we don't return it as a filter then we need to return the mapping between key and value somewhere else


richfitz (approved):
LGTM, so long as @carbowizzy is happy with the output

PR #276, opened 2021-05-11
mrc-2399: Wire up calibrate plot endpoint with naomi data preparation 

r-ash:
There are a couple of changes to note here from the dummy impl
* No longer returning columns `upper` and `lower` we just don't want to show those uncertainty ranges
* If trying to view calibration plot for a model calibrated before deployment of this feature this endpoint will error (there are changes to the calibration data structure) can we show the error & not show the graph but not block users from proceeding? So that previous model outputs can still be reviewed. Jeff is not concerned about porting previous results over, users can recalibrate to be able to see the new plot


Nicolas-Dolan:
Your code coverage is still a bit Italian:
image

EmmaLRussell (approved):
This looks pretty good to me now (See screenshot). 
Screenshot from 2021-11-08 14-54-18

PR #279, opened 2021-05-24
Generate downloads async on demand

r-ash:
This PR will
* Remove `/download/spectrum/`, `/download/coarse-output` and `/download/summary` endpoints in favour of async download generate endpoints `/download/submit/<type>/<id>` where `type` can be `summary`, `coarse-output` or `spectrum`
* Remove `uploadMetadata` from calibrate response - this will now be returned with the `download` button (note that these endpoints will need to be run before user can upload data to the ADR)

We want to do this because it will enable
* Getting translated summary report without re-calibrating
* Speed up calibration and model run (output only generated when demanded)
* Mean we can add notes in the future into the download without polluting the model calibrate endpoint
* Allow us to add options to downloads in the future too
* Remove additional data and metadata returned from calibrate which is not necessary at that point

Note this should not be merged until front end support has been added - TBC when that will be done but I expect it will probably introduce some changes


R/migrations.R
```diff
@@ -0,0 +1,102 @@
+run_migration <- function(queue, output_dir) {
```
richfitz, line 1: As discussed:

* don't write any files (I think this means you can omit `output_dir` as an argument)
* dry run mode so it's easy to test on production (i.e., don't write to redis)
* drop the first two parts to migrate_task

docker/build
```diff
@@ -29,7 +29,7 @@ ENTRYPOINT [ "/bin/bash", "-l", "-c" ]
 EOF
 
 rm -rf naomi
-git clone https://github.com/mrc-ide/naomi
+git clone --single-branch --branch mrc-2406 https://github.com/mrc-ide/naomi
```
richfitz, line 32: I've approved but not merged the naomi PR, so this will need coordinating on merge

richfitz (changes requested):

PR #286, opened 2021-07-28
mrc-2539: Add dummy endpoint for input time series graphs (shinyRob)

r-ash:
Diff looks scary but it is just 2 big test files
This PR will
* Add POST endpoint `/chart-data/input-time-series/<type>` where `type` is either `anc` or `programme` and takes a body containing files (at least `shape` and one of `anc` or `programme` but you can pass it more) and returns data formatted for plotting as input time series with filters and defaults


R/filters.R
```diff
@@ -334,3 +337,51 @@ construct_tree <- function(data, id_column = 1, parent_id_column = 2,
...
+      id = scalar("plot_type"),
+      column_id = scalar("plot"),
+      label = scalar(t_("INPUT_TIME_SERIES_FILTER_PLOT_TYPE")),
+      options = get_simple_filter(data, "plot")
+    ),
+    list(
+      id = scalar("area_level"),
+      column_id = scalar("area_level_label"),
+      label = scalar(t_("INPUT_TIME_SERIES_FILTER_AREA_LEVEL")),
+      ## TODO: Plot type and time step should use human readable name for label
```
EmmaLRussell, line 353: Did you mean to leave these TODOs? Any text fields are going to have to get translated in the front end anyway so human readability here isn't completely necessary - though it would be good if they were consistent with existing filters.

inst/schema/InputTimeSeriesResponse.schema.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "data": { "$ref": "InputTimeSeriesData.schema.json" },
+    "plottingMetadata": {
+      "type": "object",
+      "properties": {
+        "barchart": { "$ref": "BarchartMetadata.schema.json" }
```
EmmaLRussell, line 9: I guess it's good to re-use the BarchartMetadata schema here, but I don't think the prop should be called 'barchart' since the charting implementation is going to be different from the old charts - and it won't necessarily be a barchart. 
I wouldn't mind if 'plottingMetadata' just became 'metadata' and that was a BarchartMetadata object, or could stick with 'plottingMetadata' and replace 'barchart' with 'genericChart'..?

EmmaLRussell (commented):
I probably won't really know if there's any changes required until we start using it in anger, but this looks good to m!

inst/schema/InputTimeSeriesResponse.schema.json
r-ash: Oops - that is complete nonsense left from a copy + pasting. Thanks for catch

docker/test_connection
```diff
@@ -44,7 +44,7 @@ for attempt in $(seq 10); do
         echo "SUCCESS"
         exit 0
     fi
-    sleep 1
+    sleep 2
```
r-ash, line 47: Looks like it has got slower for hintr to start up - not exactly sure why but this failed 3 or 4 times in a row on CI before I increased the timings

R/filters.R
r-ash: Yeah, will fill these in with real labels when I implement the endpoint with real data. Might put code into naomi to manage that mapping of id to user readable name (this is how e.g. we are managing age group ID to human readable). 

Is doing the translations of this in the front end a new thing for this plot type? Does that mean we're going to be duping effort? Returning human readable labels elsewhere for filters is managed in hintr & naomi I thought?

EmmaLRussell: I might be misremembering, but I thought for the filters, the actual filter label, e.g. 'age', 'sex' was a translation key to be translated in the front end, but the options labels ('male' etc) were translated in hintr.

r-ash: Oh good, I think we're on the same page then. This comment is about the options. Have clarified that in the TODO.

inst/schema/InputTimeSeriesRow.schema.json
```diff
@@ -0,0 +1,36 @@
...
+      "type": "number"
+    },
+     "age_group": {
+      "type": "string"
+    },
+     "time_period": {
+      "type": "string"
+    }
+  },
+  "additionalProperties": true,
```
richfitz, line 33: Any reason this one is true?

tests/testthat/integration-server.R
```diff
@@ -134,7 +134,7 @@ test_that("validate survey", {
   expect_equal(response$data$type, "survey")
   expect_equal(response$data$fromADR, FALSE)
   expect_true(length(response$data$data) >= 20000)
-  expect_equal(typeof(response$data$data[[1]]$est), "double")
+  expect_equal(typeof(response$data$data[[1]]$estimate), "double")
```
richfitz, line 137: `testthat::expect_type` might be preferred here

richfitz (approved):
LGTM, provided that it does what the front-end needs. One query on a schema and something I noticed in a test (which is not particularly important!)

PR #290, opened 2021-08-03
Wire up time series endpoints with real data transformation functions

r-ash:


R/migrations.R
```diff
@@ -23,7 +23,7 @@ run_migration <- function(queue, output_dir, dry_run = FALSE) {
 migrate_task <- function(task_id, queue, dry_run = FALSE) {
   message(sprintf("Migrating %s", task_id))
   res <- queue$queue$task_result(task_id)
-  if (!is.null(res$version) && identical(as.character(res$version), "2.5.0")) {
+  if (!is.null(res$version) && as.character(res$version) >= "2.5.0") {
```
richfitz, line 26:
```suggestion
  if (!is.null(res$version) && numeric_version(res$version) >= "2.5.0") {
```

though curious what format the input has here?

R/time_series.R
```diff
@@ -1,5 +1,6 @@
 get_programme_time_series <- function(programme, shape) {
-  data <- read_csv(system_file("extdata/dummy_input_time_series_programme.csv"))
+  data <- naomi::prepare_input_time_series_art(programme$path, shape$path)
+  data <- as.data.frame(data)
```
richfitz, line 3:
```suggestion
  data <- as.data.frame(data, stringsAsFactors = FALSE)
```

(and elsewhere) not a huge deal, but worth adding if we don't depend on R >= 4.0

richfitz (changes requested):
tiny things, almost doesn't seem worth request changes, but the version thing is a bit off

R/migrations.R
r-ash: It should just be a character or NULL, so `as.character` cast is unnecessary

EmmaLRussell (approved):
Looks good, working perfectly with the charts, It's only getting annual data back for Antartica dataset, but I guess that's because the ART data there is all the annual (December) values.

richfitz (approved):

PR #297, opened 2021-09-22
Add barchart plotting metadata to schema

r-ash:
This has been missing for a while, now web devs want to use this metadata so we need it to be in the schema so it is available in the generated types


inst/schema/PlottingMetadataResponse.schema.json
```diff
@@ -3,7 +3,11 @@
   "definitions": {
     "choropleth_metadata": {
       "type": "array",
-      "items": { "$ref": "ChoroplethIndicatorMetadata.schema.json" }
+      "items": { "$ref": "IndicatorMetadata.schema.json" }
+    },
+    "barchart_metadata": {
```
richfitz, line 8: rob has clarified this is already included in the response, the schema allows additional properties and this was passed through but undescribed

richfitz (approved):

PR #299, opened 2021-09-24
Use separate metadata for barchart & choropleth

r-ash:
Web devs have requested this. The rename in #297 from ChoroplethIndicatorMetadata -> IndicatorMetadata has broken master for them. They would prefer to have these as separate schema.


r-ash:
Not needed, see https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1749

PR #301, opened 2021-09-24
Run codecov in paralell with tests

r-ash:
Have watched the tests running and confirmed that this is doing what we want, i.e. running code cov in parallel with all tests but not blocking the docker image push

### DIFF
--- a/buildkite/pipeline.yml
+++ b/buildkite/pipeline.yml
@@ -6,17 +6,22 @@ steps:
 
   - label: ":hammer: Unit tests"
     command: docker/test_check
+    key: "unit tests"
 
   - label: ":hammer_and_pick: Integration tests"
     command: docker/test_integration
+    key: "integration tests"
 
   - label: ":docker: Docker connection test"
     command: docker/test_connection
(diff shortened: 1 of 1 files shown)


buildkite/pipeline.yml
```diff
@@ -6,17 +6,22 @@ steps:
...
   - label: ":hammer_and_pick: Integration tests"
     command: docker/test_integration
+    key: "integration-tests"
 
   - label: ":docker: Docker connection test"
     command: docker/test_connection
+    key: "connection-tests"
 
-  - wait
+  - label: ":codecov: Codecov"
```
richfitz, line 19: is it worth trying forcing a failure here (easily done by `command: /bin/false`), to make sure we get pushed images + failed build?

r-ash: Have done that confirmed it has worked see https://buildkite.com/mrc-ide/hintr/builds/1132#_

Reverted that and force pushed to remove from history

richfitz (approved):
nice one

PR #303, opened 2021-09-28
mrc-2623: Add area and time_period metadata

r-ash:
This PR will
* Rename "filters" to "columns" in metadata - this is to better reflect what these actually are, they are metadata containing mappings of column id to label for display
* Add column metadata for time period and area - area being the hierarchy for the dataset

Uses https://github.com/mrc-ide/naomi/pull/255 and that needs to be merged in before this can be merged

Blocked on front tickets being updated


inst/schema/Column.schema.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "column_id": {
+      "type": "string"
+    },
+    "label": {
+      "type": "string"
+    },
+    "values": {
+      "oneOf": [
+        { "$ref": "NestedFilterOption.schema.json" },
```
EmmaLRussell, line 16: If `ColumnValue` is a rebrand of `FilterOption`, would it be more consistent to also have `NestedColumnValue` or similar?

EmmaLRussell (changes requested):
I think actually it would be good to retain 'area_name' in the main data after all - it's going to make the plot rendering more convoluted to have to pull that out of the column values. Sorry, should have thought of that earlier!
Otherwise, LGTM.

EmmaLRussell (approved):
Works like a charm now, thanks!

PR #310, opened 2021-11-05
Update programme test data to include vls indicators

r-ash:
This adds optional indicators to programme/ART input data
* vls_tested
* vls_suppressed
This also updates metadata so that vls_tested, vls_suppressed and art_new (a column we already had but weren't displaying) can be shown in the input graph

I've deployed this to staging and checked that it all looks OK there. Note that the indicator names not translated yet - that will come when we do the next big set of translations. Which we should do very soon.

PR #312, opened 2021-11-11
Add test for generating input time series plot when no vls data present

r-ash:
This was failing previously but has been fixed in Naomi. This should fix the build problems you were having this morning. The problem was we merged the #310 in which the input time series was assuming that the data would have new VLS columns. The test data in hint doesn't have those new columns but the columns are optional so it should work with or without them.


hillalex (approved):
Works as expected, thanks

PR #317, opened 2021-11-18
mrc-2744: Allow survey & programme validation to be run with less strict checks

r-ash:
This adds a query param `strict` to `/validate/survey-and-programme` to turn off strict checks. e.g. send `/validate/survey-and-programme?strict=false` by default `strict = TRUE`

In practice this only turns off 2 checks of ANC data namely for each row
* anc_tested - anc_tested_pos >= 0
* (anc_tested_pos + anc_known_pos) - anc_already_art >= 0

Have checked this is the desired behaviour with Jeff & Rachel.

Regarding other files, we want to make PJNZ and population optional but if they do upload them not change the checks we run. Note that as we discussed we use PJNZ to return `country` and `iso3`. Not sure how `country` is used? But `iso3` we get back from the front end when call `/meta/plotting/<iso3>`. If a user doesn't upload PJNZ if you call this endpoint with any iso3 we don't have in the graph then you will get default values. i.e. if no PJNZ you could call this with path `/meta/plotting/default` or `/meta/plotting/NULL` or something and you will get default metadata back. I could make this more formal & explicit and add a new endpoint at `/meta/plotting` to return the default values if you prefer. @EmmaLRussell let me know what you think.


EmmaLRussell (commented):
I think having a `meta/plotting` endpoint would be good for the default metadata.

Do any changes need to be made in `validate/baseline-individual`? Looks like not - it doesn't seem to be checking for missing files there.

r-ash:
> I think having a `meta/plotting` endpoint would be good for the default metadata.
> 
> Do any changes need to be made in `validate/baseline-individual`? Looks like not - it doesn't seem to be checking for missing files there.

Thanks - added that `meta/plotting` endpoint now. No changes for `validate/baseline-individual` if users upload any of those files we want to keep the same checks as we have currently.

EmmaLRussell (approved):
Looks sensible to me - haven't actually tested it in anger from HINT yet, as we're not quite there with DE mode.

PR #325, opened 2021-12-01
Add validation check that ART data is only uploaded at one area level

r-ash:


R/validation_asserts.R
```diff
@@ -312,3 +312,15 @@ assert_file_extension <- function(file, types) {
   }
   invisible(TRUE)
 }
+
+assert_single_level <- function(data) {
+  levels <- lapply(data$area_id, function(area_id) {
+    gsub("\\w{3}_(\\d+)_\\d+.*", "\\1", area_id)
```
jeffeaton, line 318: Do we use this regex elsewhere for checking?  I would not require the third block to be digits -- that will fail for South Africa; I think also maybe fails on Nigeria.

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# hintr 1.0.13
+
+* Add validation error to programme/ART data to ensure only 1 area level is uploaded
```
jeffeaton, line 3: I think this also applies to ANC data. Easy to add on that?

jeffeaton (changes requested):

R/validation_asserts.R
jeffeaton: @r-ash revised check to use `area_level` from areas dataset instead of string parsing

NEWS.md
jeffeaton: this is done

jeffeaton (approved):
thanks -- looks good

PR #328, opened 2021-12-06
Relax validation rules to check for unique area levels per year, instead of in all data

r-ash:
Will unblock Mozambique issue 39

Oli has reported that Mozambique uses ART data at multiple area levels. So we should allow this. We're changing our validation rule from rejecting ART data with multiple area levels to rejecting ART with multiple area levels per year.


jeffeaton (approved):
Thanks -- looks good to me.

PR #345, opened 2021-12-23
Trigger deployment build if running on master

r-ash:


richfitz (approved):
At some point we need to think about what happens when we move the default branch

PR #349, opened 2022-02-01
Check all columns present in programme data before using them

r-ash:
This only reduces code coverage because it removes a line

### DIFF
--- a/R/validate_inputs.R
+++ b/R/validate_inputs.R
@@ -115,7 +115,6 @@ do_validate_population <- function(population) {
 do_validate_programme <- function(programme, shape, strict = TRUE) {
   assert_file_extension(programme, "csv")
   data <- read_csv(programme$path, header = TRUE)
-  data$art_current <- as.numeric(data$art_current)
   assert_single_country(data, "programme")
   assert_column_names(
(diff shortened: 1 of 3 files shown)


R/validate_inputs.R
```diff
@@ -115,7 +115,6 @@ do_validate_population <- function(population) {
 do_validate_programme <- function(programme, shape, strict = TRUE) {
   assert_file_extension(programme, "csv")
   data <- read_csv(programme$path, header = TRUE)
-  data$art_current <- as.numeric(data$art_current)
```
richfitz, line 118: This looks like it might want shifting down the function rather than removing (i.e., after line 122). Was this previously put in to cope with things like an all-missing columns coming through as a logical NA?

richfitz: Code originally added in https://github.com/mrc-ide/hintr/commit/c2a4a863378d2778e9be6ebb344d6c342423aee3 if that helps

r-ash: Oh yeah, this is duped up. It is still on line 133 which is just before we validate the contents of that column

richfitz (approved):

PR #369, opened 2022-03-28
Add area hierarchy labels

r-ash:
This PR will
* Return `area_hierarchy` from input time series response for displaying as tooltips/labels on the input time series plots

Note that this will return data like
| area_id | area_name | area_hierarchy |
| ---------- | --------------- | -------------------- |
| MWI     | Malawi        | Malawi                |
| MWI_1  | Northern     | Northern           |
| MWI_2  | Central     | Central           |
| MWI_3  | Southern     | Southern           |
| MWI_1_1  | Northern     | Northern/Northern           |
| MWI_2_1  | Central-East     | Central/Central-East |
| MWI_2_2  | Central-West     | Central/Central-West  |
| ... | ... | ... |

It will be `NULL` for country level. Not include country level in the hierarchy and also include the current area name in the hierarchy. We can iterate on this if not perfect.

This is to support hint https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2849


EmmaLRussell (commented):
Area hierarchy data is appearing in the ITS data as described 👍 

I think it would be best if these area hierarchies matched those calculated in HINT for display in the tables, which are slightly different:
- No hierarchy shown for country level or top region level
- Current region not included in hierarchy

Though I guess to change that would not involve this PR, but would be done in Naomi?

r-ash:
That sounds good - I can update the format to match current front end format.

EmmaLRussell:
Actually, I've tried adding these to the Input Time Series subplots, and I think it makes sense to keep the current area in the hierarchy in this context, as it seems like it would be best to add it to the tooltip (hierarchies can be too long to fit in subplot area label), so it won't naturally be right next to the current area:
Screenshot from 2022-04-11 10-07-34

This means that the hierarchy will be a little different from that currently used in the ITS table, which does not include current area. These hierarchies are calculated in the front end and can continue to be so - however I've spotted an inconsistency: unlike the Map table, the ITS table shows the top (country) level. I'll add a ticket to make this consistent with Map table. 

area hier table

PR #370, opened 2022-03-31
Return warnings from programme and anc input validation

r-ash:
This PR will
* Return warnings about mismatches Naomi & Spectrum values from `/validate/survey-and-programme` endpoint for anc and programme/art data

Changes to interface
* `/validate/survey-and-programme` now takes `pjnz` in the request body which is optional, if calling endpoint for survey data this will continue as before, if for anc or art data this will error if no PJNZ file supplied
* `/validate/survey-and-programme` response will include `warnings` for art and anc data 

Hint ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3109


EmmaLRussell (approved):
> /validate/survey-and-programme now takes pjnz in the request body which is optional, if calling endpoint for survey data this will continue as before, if for anc or art data this will error if no PJNZ file supplied

Looks like we already show an error if user attempts to upload any survey and programme file before the PJNZ, and that error is thrown before it attempts to validate with hintr, so that should be fine. But it does meant we'll need to make that change in HINT to pass the PJNZ file to the endpoint before we can merge this branch. 

Out of interest, why does this not apply to survey as well as the other two, if it's checking against the shape file. Does the same logic not apply?

r-ash:
> Out of interest, why does this not apply to survey as well as the other two, if it's checking against the shape file. Does the same logic not apply?

Only because we've not had an example from Jeff & Rachel of a warning they want to return from survey upload. Though actually if we did support that it would simplify some of this logic quite a bit and possibly future proof it. Is it much extra work on your end to include warnings from survey upload?

Nicolas-Dolan (approved):

PR #379, opened 2022-05-18
mrc-3130: Allow (optional) body for notes on download submit endpoint

r-ash:
This will add the notes into the output zip as `notes.txt`. I'm hoping this shouldn't affect the current working of hint, the notes are optional part of the body. Needs to be merged after https://github.com/mrc-ide/naomi/pull/319 and replaces #378


inst/schema/DownloadSubmitRequest.schema.json
```diff
@@ -0,0 +1,29 @@
...
+      "required": [ "name", "updated", "note" ]
+    }
+  },
+  "type": "object",
+  "properties": {
+    "notes" : {
+      "type": "object",
+      "properties": {
+        "project_notes": { "$ref": "#/definitions/note" },
+        "version_notes": {
```
EmmaLRussell, line 21: Is this idea here that the notes posted should include version notes from this and all previous versions? That's a little tricky since, as we've discussed, we don't really have the notion of version parentage..

inst/schema/DownloadSubmitRequest.schema.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "definitions": {
+    "note": {
+      "type": "object",
+      "properties": {
+        "name": { "type": "string" },
+        "updated": { "type": "string" },
```
EmmaLRussell, line 8: Is this meant to be the updated timestamp for the note or the project/version? We have the latter, not the former.

EmmaLRussell (commented):
This seems to work fine with current hint not sending any notes. Just a couple of questions.

inst/schema/DownloadSubmitRequest.schema.json
r-ash: Was thinking from this version and all other versions of this project, show it in date order. We don't have the trail of parentage but can we get this? Looks like db has `project_version` table so would we be able to get all `project_version` notes with this `project_id`?

inst/schema/DownloadSubmitRequest.schema.json
r-ash: Yeah. Was thinking last updated time, where the last updated time for the project is the the latest time from all the different versions. Assuming that is what we show in the UI? I am thinking this should just match the times we show in the UI
Screenshot_20220519_150500

I meant to ask - what format should we return here? Is a string sensible?

inst/schema/DownloadSubmitRequest.schema.json
EmmaLRussell: Yes, we could use version updated date to order the versions. Would you want to include later versions though, if this isn't the latest? In that case, should we highlight which version is this version.

r-ash: Good question - I will ask raise with Jeff/Rachel and make a note in the front end ticket

EmmaLRussell (approved):

PR #380, opened 2022-05-20
mrc-3127: Download submit can take (optional) project state

r-ash:
This is a WIP, couple of failing tests remaining

PR #384, opened 2022-06-07
mrc-3116: Add async rehydrate endpoints to return state to front end 

r-ash:
This PR builds on #380 and should be merged after that

This PR adds 3 endpoints
* `/rehydrate/submit` - which takes path to output zip and returns ID of async rehydrate job
* `/rehydrate/status/<id>` - gets status of job (uses same format as model fit, calibrate, and download endpoints in response) note that progress here is empty for now but we could add some progress messages in the future when this becomes more detailed process (see below)
* `/rehydrate/result/<id>` - gets the result of the job (this returns the JSON required to rehydrate the state in the front end)

After this has been merged next steps for the API are
* Provide good testing setup for hint, we want an output zip available which they can use to recover some state but work with them to decide what would be most useful on their end. I am thinking some script which can be run after starting the API to run a model and save output zip with state json included
* Add backwards compatibility handling to this, my rough plan is have some migration code which looks at the version number in the JSON in the uploaded zip and runs through migration to current if needed. This might do things like add a new row to the data and save a new copy of it to disk and use new path in returned state. Add a value for a new model option and return this in the JSON state to the front end
* Add rehydration for files without JSON state (e.g. if something saved by Jeff in R which doesn't have existing state info in it) will have to read input from the zip, save as input files to disk, create a model output and save to redis (etc.) then construct the state JSON and return to front end


EmmaLRussell (approved):
Looks good to me

PR #385, opened 2022-06-07
Switch download submit to a POST endpoint

r-ash:
This endpoint should probably have been a `POST` from the start but now that it takes a json body it needs to be switched

### DIFF
--- a/R/api.R
+++ b/R/api.R
@@ -311,7 +311,7 @@ endpoint_download_submit <- function(queue) {
                                                 schema_root())
   response <- porcelain::porcelain_returning_json(
     "DownloadSubmitResponse.schema", schema_root())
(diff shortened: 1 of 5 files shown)


LekanAnni (approved):
LGTM 👍

EmmaLRussell (approved):
LGTM but I guess we shouldn't merge this until HINT has been updated to use POST instead of GET - do we have a ticket for that? Would that be the one for pushing project state to this endpoint?

LekanAnni:
> LGTM but I guess we shouldn't merge this until HINT has been updated to use POST instead of GET - do we have a ticket for that? Would that be the one for pushing project state to this endpoint?

That's right, I am working on the POST update via this branch, so we shouldn't merge yet.

PR #386, opened 2022-06-09
Remove download from state

r-ash:
Builds on #385 

We don't necessarily know the download ids at the point of generation (and we definitely don't know the ID of the output zip download ID) see https://github.com/mrc-ide/hint/pull/708#pullrequestreview-1001523167

### DIFF
--- a/inst/schema/ProjectState.schema.json
+++ b/inst/schema/ProjectState.schema.json
@@ -18,14 +18,6 @@
       },
       "additionalProperties": false,
       "required": [ "options", "id" ]
-    },
-    "async_run": {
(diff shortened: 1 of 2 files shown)


LekanAnni:
> Builds on #385
> 
> We don't necessarily know the download ids at the point of generation (and we definitely don't know the ID of the output zip download ID) see [mrc-ide/hint#708 (review)](https://github.com/mrc-ide/hint/pull/708#pullrequestreview-1001523167)

I think it should target post-endpoint instead of master

LekanAnni (approved):
LGTM

PR #388, opened 2022-07-01
Save project state into subdirectory of output zip

r-ash:
This PR will
* Move the project state into a subdirectory as requested by @jeffeaton

PR #389, opened 2022-07-12
Run integration tests in a new environment with pkload env as parent

r-ash:
This PR will
* Fix integration test setup after pkgload 1.3.0 update see https://pkgload.r-lib.org/news/index.html

> The loaded namespace is now locked just before user onLoad hooks are run. This better reproduced the namespace sealing behaviour of regular loading.
The package environment environment is now locked as well before both the user and package onAttach hooks are run.


r-ash:
Not really sure what is causing the unit test failures on #388 but must be something I have changed as part of that PR

richfitz (approved):

PR #390, opened 2022-07-13
Move project state into subdirectory

r-ash:
This PR will
* Move the project state into a subdirectory as requested by @jeffeaton


R/endpoints.R
```diff
@@ -120,19 +123,6 @@ input_time_series <- function(type, input) {
   })
 }
 
-model_options <- function(input) {
```
r-ash, line 123: There is removing a dupe function definition

richfitz (approved):

PR #393, opened 2022-08-11
mrc-3215: Set country specific default values for all model and calibrate options

r-ash:
This PR will
* Get country specific hardcoded defaults from naomi
* Merge these with the defaults from the data (we want to prefer hardcoded value but fallback to data default if e.g. the hardcoded value is not in the set of possible options)
* Return to the front end
* Use these country specific defaults for calibration options too

Changes for front end
* `/calibrate/options` endpoint now takes iso3 in the path `/calibrate/options/<iso3>` so that we can get country specific default options
* The options schema has changed so that for a `multiselect` control the value can either be a single string or an array of strings. This is because we now has examples of defaults we want to set with 2 options chosen whereas previously we were only supporting 1 default value


r-ash:
Replaced this with #405

PR #397, opened 2022-08-19
Integrate data for comparison plot

r-ash:
This PR will
* Replace comparison plot mock data with real data from naomi

Note if a user tries to get comparison plot data on an old model fit then hintr will return an error prompting user to re-run their fit. Can we show this error if they try to view the plot?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,7 @@ Imports:
     ids,
     jsonlite (>= 1.2.2),
     lgr,
-    naomi (>= 2.7.0),
+    naomi (>= 2.7.1),
     porcelain (>= 0.1.8),
     R6,
     redux,
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -240,20 +240,19 @@ calibrate_plot <- function(queue) {
 comparison_plot <- function(queue) {
   function(id) {
     verify_result_available(queue, id)
-    data <- readRDS(system_file("comparison_plot_mock.rds"))
-    data <- data[c("area_id", "area_name", "age_group", "sex",
-                   "calendar_quarter", "indicator", "source", "mean",
-                   "lower", "upper")]
+    data <- naomi::hintr_comparison_plot(queue$result(id))
     ## Strip tibble class to work with helper functions which rely on
     ## converting to vector when selecting 1 column
+    data <- as.data.frame(data)
+    data <- data[, c("area_id", "area_name", "age_group", "sex",
+                   "calendar_quarter", "indicator", "source", "mean",
+                   "lower", "upper")]
     filters <- get_comparison_plot_filters(data)
     list(
       data = data,
       plottingMetadata = list(
         barchart = list(
-          ## TODO: add and use calibrate barchart metadata (or make it clear that
-          ## we want to use the same here)
-          indicators = get_barchart_metadata(data, "calibrate"),
+          indicators = get_barchart_metadata(data, "comparison"),
           filters = filters,
           defaults = get_comparison_barchart_defaults(data, filters)
         )
--- a/docker/build
+++ b/docker/build
@@ -29,7 +29,7 @@ ENTRYPOINT [ "/bin/bash", "-l", "-c" ]
 EOF
 
 rm -rf naomi
-git clone --single-branch --branch return-data-inputs https://github.com/mrc-ide/naomi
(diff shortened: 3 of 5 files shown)


EmmaLRussell (commented):
> This PR will

>    Replace comparison plot mock data with real data from naomi

So when this is merged, we can remove the feature switch for showing the comparison plot?

> Note if a user tries to get comparison plot data on an old model fit then hintr will return an error prompting user to re-run their fit. Can we show this error if they try to view the plot?

As far as I know, when an error is returned by the calibrate plot endpoint, it is committed to the store, but I don't think we currently display it anywhere. @Nicolas-Dolan could that be added, in your show barchart branch (or on a separate branch)?

This looks pretty good to me. I ran this up with the barchart branch, and could see data displayed. The only weird thing was that, using the Antarctica dataset, September 2018 was listed as a Period, but I couldn't seem to get any data showing no matter what combination of filters I tried. 
image
There is September 2018 data showing in the Map, Bubble chart, and estimates-only Barchart. And, actually, those have June 2019 data showing too, which doesn't seem to be available in the Comparison barchart. Should it be?

r-ash:
> > This PR will
> 
> > Replace comparison plot mock data with real data from naomi
> 
> So when this is merged, we can remove the feature switch for showing the comparison plot?
> 
> > Note if a user tries to get comparison plot data on an old model fit then hintr will return an error prompting user to re-run their fit. Can we show this error if they try to view the plot?
> 
> As far as I know, when an error is returned by the calibrate plot endpoint, it is committed to the store, but I don't think we currently display it anywhere. @Nicolas-Dolan could that be added, in your show barchart branch (or on a separate branch)?
> 
> This looks pretty good to me. I ran this up with the barchart branch, and could see data displayed. The only weird thing was that, using the Antarctica dataset, September 2018 was listed as a Period, but I couldn't seem to get any data showing no matter what combination of filters I tried. image There is September 2018 data showing in the Map, Bubble chart, and estimates-only Barchart. And, actually, those have June 2019 data showing too, which doesn't seem to be available in the Comparison barchart. Should it be?

I've checked this with in the app and we can 2018 for some combination. Specifically indicator either ART number (attending) or ANC ART coverage age matched. Area Malawi, period Sept 2018 and 2016, sex both age 0-14 or  15+. I've checked with Rachel and she says the data will be more spare re only data for a small set of diaggregates.

"My script will have aligned the data so we only have disags available where there is both input and output data. ART data comes in aggregated so we can only compare it to aggregated model outputs. ANC will likely only have a single age/sex combo per area_id."

There is no 2019 in the comparison plot because 2019 in the forward projection time point. This plot shows inputs and output comparison and so can only compare model outputs to matched data inputs. Not the projection time point.

Nicolas-Dolan:
> I think because the data will be quite sparse we should be put the standard "No data available for these selections" message that we have on the input plots. That sound ok @Nicolas-Dolan ?

Sure thing, I'll integrate an error message into the show plot branch.

rtesra (commented):
@Nicolas-Dolan @r-ash I think the idea was to have the the filter populated by what was available in the dataframe? So only filters that are present in the dataframe would be available in the drop down?

r-ash:
> @Nicolas-Dolan @r-ash I think the idea was to have the the filter populated by what was available in the dataframe? So only filters that are present in the dataframe would be available in the drop down?

Yes that is the case but there will still be combinations of filters which lead to no data being available. e.g. in example above ART number (attending) and age group 15-49 would give this.

Nicolas-Dolan:
> As far as I know, when an error is returned by the calibrate plot endpoint, it is committed to the store, but I don't think we currently display it anywhere. @Nicolas-Dolan could that be added, in your show barchart branch (or on a separate branch)?

I've written up a ticket for this: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3559/Display-any-errors-saved-to-modelCalibrate-store

> I think because the data will be quite sparse we should be put the standard "No data available for these selections" message that we have on the input plots. That sound ok @Nicolas-Dolan ?

and another ticket for this: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3558/Display-no-data-are-available-message-in-barchart

Nicolas-Dolan (approved):
All worked fine when I targeted this hintr branch with the comparison chart pr.

PR #406, opened 2022-09-20
Naomi return more informative error message

r-ash:


R/model_options.R
```diff
@@ -188,19 +188,6 @@ build_options_from_template <- function(options_template) {
...
-#' @param options Key value list of model options
-#'
-#' @return TRUE if valid else throws an error
-#' @keywords internal
-do_validate_model_options <- function(data, options) {
-  list(
-    valid = scalar(naomi:::validate_model_options(data, options))
-  )
-}
-
```
r-ash, line 203: Just cleaning up this stale function as I noticed it was sitting around. Nothing to do with this PR

PR #411, opened 2022-09-29
mrc-3655: Replace rehydate with load in error message wording for clarity

r-ash:
UNAIDS were confused about "rehydrate" and we call it "load" elsewhere in the UI so rename it here to be consistent


LekanAnni (approved):
LGTM 👍

PR #422, opened 2022-11-17
Improve performance of get_model_options

r-ash:
This will speed up in a few ways
* Reduce duplicate reads (of shape file and geojson)
* Use vectorised version of naomi functions for getting labels (atm this is slow because each call to naomi translates some strings, now that translation is only done once)
* Use data.table to read the csv file, this much faster than read.csv

Running benchmarking locally before these changes on current Cameroon data from ADR we have
```
> microbenchmark::microbenchmark(model_options(input), times = 10)
Unit: seconds
                 expr     min       lq     mean   median       uq      max neval
 model_options(input) 2.14557 2.154384 2.249949 2.232277 2.351745 2.373982    1
```

After this we are getting

```
> microbenchmark::microbenchmark(model_options(input), times = 10)
Unit: milliseconds
                 expr      min       lq    mean   median      uq      max neval
 model_options(input) 376.0353 383.2105 406.114 385.1356 395.946 561.7291    10
```

Looking at the profile the timing is now
* reading geojson `hintr_geojson_read` ~ 120ms
* getting filters from this  `get_region_filters` ~ 50ms
* reading the survey data `read_csv` ~ 40ms
* reading anc data `naomi::read_anc_testing` ~ 20ms
* building the options json `naomi.options::get_controls_json` ~ 80ms

I think this will be a bit quicker live because we have caching of the geojson read.

### DIFF
--- a/R/read_data.R
+++ b/R/read_data.R
@@ -24,16 +24,12 @@ read_csv_regions <- function(csv_file) {
 }
 
 read_csv <- function(file, ...) {
-  header <- readLines(file, 1)
-  if (!grepl(",", header) && grepl(";", header)) {
-    read <- utils::read.csv2
-  } else {
-    read <- utils::read.csv
-  }
-
-  data <- read(file, ..., stringsAsFactors = FALSE)
-  na_or_empty_rows <- rowSums(is.na(data)) + rowSums(data == "", na.rm = TRUE)
-  data[na_or_empty_rows != ncol(data), ]
+  data <- data.table::fread(file, ...,
+                            blank.lines.skip = TRUE,
+                            data.table = FALSE,
+                            nThread = 1,
+                            na.strings = c("NA", ""))
+  data[rowSums(is.na(data)) != ncol(data), ]
 }
 
 read_pjnz_iso3 <- function(pjnz) {
@@ -56,6 +52,10 @@ read_geojson_iso3 <- function(shape) {
   json <- hintr_geojson_read(shape)
   ## At this point we have validated there is only data for 1 country so we can
   ## just take the first.
+  get_geojson_iso3(json)
+}
+
+get_geojson_iso3 <- function(json) {
   substr(json$features[[1]]$properties$area_id, 1, 3)
 }
 
@@ -71,4 +71,3 @@ read_country <- function(pjnz_path) {
   hiv_params <- specio::read_pjn_metadata(pjnz_path)
   hiv_params$country
 }
-
--- a/docker/build
+++ b/docker/build
@@ -29,7 +29,7 @@ ENTRYPOINT [ "/bin/bash", "-l", "-c" ]
 EOF
 
 rm -rf naomi
-git clone https://github.com/mrc-ide/naomi
+git clone --single-branch --branch vectorise-quarter https://github.com/mrc-ide/naomi
 if [ -z $NAOMI_SHA ]; then
     git -C $PACKAGE_ROOT/naomi checkout $NAOMI_SHA
 fi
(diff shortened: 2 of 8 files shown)


jeffeaton:
Comments on this:

* Is `readr::read_csv()` sufficiently performant to use instead of `data.table::fread()`? `readr` is back-ended by `vroom` now, and I think competitive with `data.table`.
* Would it be appropriate to replace some of the reads with the Naomi read functions? E.g. for reading survey data set: https://github.com/mrc-ide/naomi/blob/45f37ef23cd600fe293b07df170c19124248b4d4/R/read-data.R#L40

r-ash:
> * Is `readr::read_csv()` sufficiently performant to use instead of `data.table::fread()`? `readr` is back-ended by `vroom` now, and I think competitive with `data.table`.

```
> microbenchmark::microbenchmark(readr::read_csv(survey), data.table::fread(survey, nThread = 1), vroom::vroom(survey), times = 10)
```
So for the survey file it is quicker

Screenshot_20221118_120436

> * Would it be appropriate to replace some of the reads with the Naomi read functions? E.g. for reading survey data set: https://github.com/mrc-ide/naomi/blob/45f37ef23cd600fe293b07df170c19124248b4d4/R/read-data.R#L40

Ideally yes it would be best if we had a function which reads each file type and we use this everywhere we read, but I'm aware some different behaviour of data reading between naomi and hintr. I thought it would be quite fiddly and slow to make sure that is all consistent and works for both hintr and hint for all input data. Though on quick look it doesn't look like much difference in survey reading at least. I think my preference would be to go with this for now, and we can try tidy up reading later when there is less time pressure

R/read_data.R
```diff
@@ -24,16 +24,12 @@ read_csv_regions <- function(csv_file) {
...
-
-  data <- read(file, ..., stringsAsFactors = FALSE)
-  na_or_empty_rows <- rowSums(is.na(data)) + rowSums(data == "", na.rm = TRUE)
-  data[na_or_empty_rows != ncol(data), ]
+  data <- data.table::fread(file, ...,
+                            blank.lines.skip = TRUE,
+                            data.table = FALSE,
+                            nThread = 1,
+                            na.strings = c("NA", ""))
+  data[rowSums(is.na(data)) != ncol(data), ]
```
richfitz, line 32: `data[!complete.cases(data), ]` I think is the canonical way of doing this

docker/build
```diff
@@ -29,7 +29,7 @@ ENTRYPOINT [ "/bin/bash", "-l", "-c" ]
 EOF
 
 rm -rf naomi
-git clone https://github.com/mrc-ide/naomi
+git clone --single-branch --branch vectorise-quarter https://github.com/mrc-ide/naomi
```
richfitz, line 32: don't forget to revert

R/read_data.R
```diff
@@ -24,16 +24,12 @@ read_csv_regions <- function(csv_file) {
 }
 
 read_csv <- function(file, ...) {
-  header <- readLines(file, 1)
-  if (!grepl(",", header) && grepl(";", header)) {
```
richfitz, line 28: does data.table cope with both delimiters ok? do we have tests for this? I presume so...

richfitz (changes requested):

R/read_data.R
richfitz: oh sorry, I misread the condition...

R/read_data.R
r-ash: It does, we have a test https://github.com/mrc-ide/hintr/blob/master/tests/testthat/test-read-data.R#L71

richfitz: fab thanks

PR #424, opened 2022-11-23
Error immediately if comparison report cannot be generated

r-ash:
At the moment naomi checks what version of the model output it can generate the comparison report for. This means that  if they try to generate it for an old model output, the download starts and the error is thrown by naomi when running on the worker. So only surfaces when the result is then fetched from hint. We are not handling errors in hint from download result atm so if we error earlier when we try to submit it can bypass this problem. We do need to add this error handling to hint in the future but this should buy us some time before the workshop.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hintr
 Title: R API for calling naomi district level HIV model
-Version: 1.0.43
+Version: 1.0.44
 Authors@R: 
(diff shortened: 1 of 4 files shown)


r-ash:
I've deployed this on dev and it results in the button being greyed out and error message being shown
Screenshot_20221123_164504

EmmaLRussell (approved):
Confirmed this woks in local dev too

PR #429, opened 2022-11-29
Add default selection data into comparison plot response

r-ash:
This will add placeholder multiple defaults for comparison plot, these will be used to switch the default filter selections shown when the user changes the indicator

### DIFF
--- a/R/filters.R
+++ b/R/filters.R
@@ -240,7 +240,7 @@ get_comparison_barchart_selections <- function(output, filters) {
   all_sexes <- get_selected_mappings(filters, "sex")
   both <- get_selected_mappings(filters, "sex", "both")
   female <- get_selected_mappings(filters, "sex", "female")
-  first_quarter <- get_selected_mappings(filters, "quarter")[1]
+  survey_quarter <- get_selected_mappings(filters, "quarter")[2]
   source = get_selected_mappings(filters, "source")
   fifteen_to_49 <- get_selected_mappings(filters, "age", "Y015_049")
   list(
@@ -250,7 +250,7 @@ get_comparison_barchart_selections <- function(output, filters) {
       disaggregate_by_id = scalar("source"),
       selected_filter_options = list(
         area = area_default,
-        quarter = first_quarter,
+        quarter = survey_quarter,
         sex = both,
         age = five_year_age_groups,
         source = source
@@ -262,7 +262,7 @@ get_comparison_barchart_selections <- function(output, filters) {
       disaggregate_by_id = scalar("source"),
       selected_filter_options = list(
         area = area_default,
-        quarter = first_quarter,
+        quarter = survey_quarter,
         sex = both,
         age = five_year_age_groups,
         source = source
@@ -274,7 +274,7 @@ get_comparison_barchart_selections <- function(output, filters) {
       disaggregate_by_id = scalar("source"),
       selected_filter_options = list(
         area = area_default,
-        quarter = first_quarter,
+        quarter = survey_quarter,
         sex = all_sexes,
         age = get_selected_mappings(filters, "age", "Y015_999"),
         source = source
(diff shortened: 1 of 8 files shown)


R/filters.R
```diff
@@ -234,17 +234,70 @@ get_calibrate_barchart_defaults <- function(filters) {
...
-      area = get_area_level_filter_option(output),
-      quarter = get_selected_mappings(filters, "quarter")[2],
-      sex = get_selected_mappings(filters, "sex")[1],
-      age = get_selected_mappings(filters, "age",
-                                  naomi::get_five_year_age_groups()),
-      source = get_selected_mappings(filters, "source")
+    list(
+      indicator_id = scalar("prevalence"),
+      x_axis_id = scalar("age"),
+      disaggregate_by_id = scalar("source"),
```
EmmaLRussell, line 242: I'd be surprised if these need to vary per indicator? But I guess this way we can keep our options open.

R/filters.R
```diff
@@ -234,17 +234,70 @@ get_calibrate_barchart_defaults <- function(filters) {
...
-    x_axis_id = scalar("age"),
-    disaggregate_by_id = scalar("source"),
-    selected_filter_options = list(
-      area = get_area_level_filter_option(output),
-      quarter = get_selected_mappings(filters, "quarter")[2],
-      sex = get_selected_mappings(filters, "sex")[1],
-      age = get_selected_mappings(filters, "age",
-                                  naomi::get_five_year_age_groups()),
-      source = get_selected_mappings(filters, "source")
+    list(
```
EmmaLRussell, line 239: Don't we also need to retain the indicator to select by default?

r-ash: I've proposed we just use the first see https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3843/Improve-comparison-plot-usability#focus=Comments-97-18340.0-0, but open to making that more explicit

R/filters.R
r-ash: Yeah, so `disaggregate_by_id` is special here as we don't show that as a filter for users. But I think possible they might change `x_axis_id`. So I think keeping them configurable makes sense

R/filters.R
r-ash: Ok so sounds like this will mean we cannot specify the default indicator separately, it will always be the first. We could do something like

```
"barchart": {
  "defaultSelections": [
     {..}
  ],
  "defaultIndicator": "prevalence"
}
```

Could even make default selections an object and have
```
{
  "prevalence": {...},
  "art_current": {...},
  ...
}
```

LekanAnni:
If we will have an ordered defaults array, we can just choose the first index but we might not. In that case we should have default prevalence.

inst/schema/ComparisonBarchartMetadata.schema.json
```diff
@@ -10,11 +10,12 @@
       "type": "array",
       "items": { "$ref": "Filter.schema.json" }
     },
-    "defaults": {
+    "defaults": { "$ref": "BarchartDefaults.schema.json" },
```
EmmaLRussell, line 13: OK, this seems a bit weird to me, but it would certainly include all the metadata you need to set the default. But doesn't this mean you're sending the full default selections for the default indicator twice, because they'll be in the `selections` array as well? Why not just send `defaultIndicator` as a string, and let HINT pick out the corresponding selections from the array (or use the first in the list if not provided? Don't feel particularly strongly about this as amount of data is low, but surprised at the repetition implied here.

EmmaLRussell: Oh, actually, is the reason so that HINT can just keep using `defaults` in the same way as it does now, directly from the response, with no code changes at all? And then to change indicators, it could actually edit the response object in place by swapping in the newly selected indicator's selections? That would be even less code to change than my 'facade' suggestion. I never really like editing the response object (can make things harder to debug), but that would certainly mean the change in HINT would be minimal.

LekanAnni (changes requested):
art_attending should have default selection as well.

inst/schema/ComparisonBarchartMetadata.schema.json
r-ash: That sounds good to me, agree funny to dupe data but it isn't a huge amount. I think in mind of time lets go with this for now as Lekan has his branch working with it. Let's make sure we review when we come to doing the full filtering of filters/default selections for all plots PR next year

PR #431, opened 2022-11-30
Make options take > 2mins to respond to test haproxy loadbalancing

r-ash:


r-ash:
Used this to test the appropriate load algorithm, not needed anymore

PR #437, opened 2022-12-08
Remove stack trace from being shown to users

r-ash:
This PR will
* Update default fallback value for anc year to 2022
* Remove stack trace from being shown to users by removing `trace` property from errors. Instead returns a `job_id`


r-ash:
image

r-ash:
Have made some further changes since last PR which will do this properly by removing the `trace` from being returned at all and instead return a `job_id`

richfitz (commented):
I thought our original idea here (reaching back through the sands of time) was that we'd make the trace hidden by default in the interface? Is that now not even what is wanted?

r-ash:
> I thought our original idea here (reaching back through the sands of time) was that we'd make the trace hidden by default in the interface? Is that now not even what is wanted?

Trace atm is hidden by default, users have to click the arrow to expand and see the whole trace. But no, now we can get the trace easily from the debug request we don't want to show it to users. This PR with https://github.com/mrc-ide/hint/pull/802 will instead show job ID always. So an error will look like

No ART data found for quarter CY2022Q3
**Job ID**: ff12313456788

richfitz (approved):

PR #442, opened 2022-12-13
mrc-3885: Add endpoint for uploading input file

r-ash:
This builds on #441 

This PR will
* Add an endpoint which will allow an input file to be uploaded

### DIFF
--- a/R/api.R
+++ b/R/api.R
@@ -373,7 +373,7 @@ endpoint_upload_input <- function(queue) {
   response <- porcelain::porcelain_returning_json(
     "SessionFile.schema", schema_root())
   porcelain::porcelain_endpoint$new("POST",
-                                    "/upload/input/<filename>",
+                                    "internal/upload/input/<filename>",
                                     upload_input(queue),
                                     input,
                                     returning = response)
--- a/R/api.R
+++ b/R/api.R
@@ -28,6 +28,7 @@ api_build <- function(queue, validate = FALSE, logger = NULL) {
   api$handle(endpoint_rehydrate_submit(queue))
   api$handle(endpoint_rehydrate_status(queue))
   api$handle(endpoint_rehydrate_result(queue))
+  api$handle(endpoint_upload_input(queue))
   api$handle(endpoint_hintr_version())
   api$handle(endpoint_hintr_worker_status(queue))
   api$handle(endpoint_hintr_stop(queue))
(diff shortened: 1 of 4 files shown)


R/upload.R
```diff
@@ -0,0 +1,18 @@
+upload_input <- function(queue) {
+  function(file, filename) {
+    md5 <- toupper(digest::digest(file, algo = "md5"))
+    dest_path <- file.path(
+      queue$uploads_dir,
+      paste0(md5, ".", tools::file_ext(filename)))
+    if (!file.exists(dest_path)) {
+      t <- tempfile()
+      writeBin(file, t)
+      file.copy(t, dest_path)
```
r-ash, line 10: This will be safe with multiple hintr instances running?

richfitz: it depends what you're trying to do. If you want an atomic write, then you need to make the tempfile be on the same filesystem as the destination, then use `file.rename`. So if `dest_path` is your upload directory, we could do

```
dest_path <- file.path(
  queue$uploads_dir,
  paste0(md5, ".", tools::file_ext(filename)))
if (!file.exists(dest_path)) {
  tmpdir <- file.path(queue$uploads_dir, ".incoming"))
  dir.create(tmpdir, FALSE, TRUE
  t <- tempfile(tmpdir = tmpdir)
  writeBin(file, t)
  file.rename(t, dest_path)
}
```

as you have it, you might as well do `writeBin(file, dest_path)` as that will be at least as fast as the later `file.copy`

Of course,  if the other copies of hint/hintr are doing a plain write you have a potential race condition, but tbh it's not that likely

R/api.R
```diff
@@ -366,6 +367,18 @@ endpoint_adr_metadata <- function(queue) {
                                     returning = response)
 }
 
+
+endpoint_upload_input <- function(queue) {
+  input <- porcelain::porcelain_input_body_binary("file")
+  response <- porcelain::porcelain_returning_json(
+    "SessionFile.schema", schema_root())
+  porcelain::porcelain_endpoint$new("POST",
+                                    "/upload/input/<filename>",
```
richfitz, line 376: I wonder about putting this below some other top level path (internal/upload perhaps) if there are many to make it clear it's not really part of the public API?

R/upload.R
r-ash: Thanks, though not likely would be good to avoid it. Have added this, thanks!

R/api.R
r-ash: Yeah sounds like a good idea

richfitz (approved):

PR #444, opened 2022-12-14
mrc-3888: Add prerun endpoint

r-ash:
This builds on #443 

This endpoint will
* Take details of input and output files, check that they exist on disk (will error if they do not). Create an ID for outputs and put into redis. Then format response as project state JSON

PR #445, opened 2022-12-14
mrc-3891: Remove old prerun functions

r-ash:
Builds on #444 

We're going to replace prerun with a more transparent version where users will recover a prerun output by uploading an output zip. Pre-emptively removing this to avoid confusion with name clashes.

Note that I have left `--prerun-dir` in main API. I'm keen to try and get this prerun in for workshop tomorrow and want to punt on removing this properly from the deploy tool (any removing the volumes from deployed apps) until I have a bit more time.


R/main.R
```diff
@@ -21,15 +21,13 @@ Options:
...
-       prerun_dir = validate_path(dat[["prerun_dir"]]),
        inputs_dir = validate_path(dat[["inputs_dir"]]))
 }
 
 main_api <- function(args = commandArgs(TRUE)) {
   # nocov start
   dat <- main_api_args(args)
-  api <- api(dat$queue_id, dat$workers, dat$results_dir, dat$prerun_dir,
-             dat$inputs_dir)
+  api <- api(dat$queue_id, dat$workers, dat$results_dir, dat$inputs_dir)
```
richfitz, line 30: above in the docopt string, should prerun-dir go (with deletions also in the tests aftwards, possibly affecting deploy?)

richfitz (changes requested):
one possible extra deletion to do

richfitz (approved):

PR #446, opened 2022-12-14
mrc-3889: Add hintr prerun function

r-ash:
This PR will
* Add a new function which takes input files, model fit output, model calibrate output. Uploads these onto the specified server using new endpoints. Then creates the output zip including the state JSON and returns the path to that file to the user. This file can then be uploaded into the app to restore a project.


R/prerun.R
```diff
@@ -1,3 +1,85 @@
...
+#' @param server The server URL to upload files to
+#' @param port The port the API is running on
+#' @param output_zip_path The path to save the output zip at, will use
+#'   a tempfile by default
+#'
+#' @return Path to the generated output zip
+#' @export
+hintr_prerun <- function(inputs, model_output, calibrate_output,
+                         server = "https://naomi.unaids.org",
+                         port = "8888",
```
richfitz, line 24: Make clear in the docs that this requires VPN/RDP access?

R/prerun.R
```diff
@@ -1,3 +1,85 @@
...
+
+  if (!is.null(port)) {
+    url <- paste0(server, ":", port)
+  } else {
+    url <- server
+  }
+
+  uploaded_inputs <- lapply(inputs, function(input) {
+    filename <- basename(input)
+    res <- httr::POST(paste0(url, "/internal/upload/input/", filename),
```
richfitz, line 45: do we have a way of querying to see which of these are needed to be uploaded? or I guess if this is only ever on the imperial network that's not really a big problem...

R/prerun.R
```diff
@@ -1,3 +1,85 @@
...
+  if (!naomi:::is_hintr_output(calibrate_output)) {
+    stop("Calibrate output must be hintr_output object")
+  }
+  required <- c("pjnz", "shape", "population", "survey")
+  optional <- c("programme", "anc")
+  assert_names(inputs, required, optional)
+  assert_files_exist(inputs)
+
+  if (!is.null(port)) {
+    url <- paste0(server, ":", port)
```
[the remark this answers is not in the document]
r-ash: Actually no, added this only for testing. Wanted to hit `example.com` with no port specified to test a request not working but

R/prerun.R
r-ash: Not at the moment, there is a ticket for that https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3887/Add-endpoint-to-check-if-files-exist would be good to do.

inst/schema/File.schema.json
```diff
@@ -5,6 +5,6 @@
     "path": { "$ref": "FilePath.schema.json" },
     "filename": { "type": "string" }
   },
-  "additionalProperties": false,
+  "additionalProperties": true,
```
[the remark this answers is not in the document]
r-ash: The response from `upload` endpoints returns `path`, `filename` and `hash` which is SessionFile object. 
The input to `prerun` endpoint only needs `path` and `filename` as it builds the `hash` from the contents of the file.

But in this context after uploading it is sending output from `upload` endpoint to `prerun` and so has a `hash` included, so I updated this schema to take additional properties.

richfitz (approved):

PR #447, opened 2023-01-04
Run tests separately from R CMD check for better output

r-ash:
This PR will update how tests are run on CI so that we can
* See time for each file
* See all failures at the same time
* Run in parallel (though I expect we need to run this on an agent with more cores for this to speed things up)


R/endpoints.R
```diff
@@ -32,12 +32,16 @@ model_options <- function(input) {
...
   tryCatch({
     json_verbatim(
-      naomi.options::get_controls_json("calibration", iso3, NULL, NULL))
+      get_controls_json("calibration", iso3, NULL, NULL))
   }, error = function(e) {
     hintr_error(e$message, "INVALID_CALIBRATION_OPTIONS")
   })
 }
 
+get_controls_json <- function(...) {
```
richfitz, line 41: What's the idea with this wrapper? You can also achieve this by doing

```
#' @importFrom naomi.options get_controls_json
```

if you just want to drop the `::` for some reason

R/zzz.R
```diff
@@ -29,7 +29,7 @@ hintr_init_traduire <- function() {
   namespaces <- "translation"
   traduire::translator_register(resources = NULL,
                                 language = languages[[1]],
-                                default_namspace = namespaces[[1]],
+                                default_namespace = namespaces[[1]],
```
richfitz, line 32: oh dear. How was this not caught? does the line not actually do anything?

buildkite/pipeline.yml
```diff
@@ -18,6 +20,8 @@ steps:
 
   - label: ":codecov: Codecov"
     command: docker/coverage
+    agents:
+      queue: "parallel"
```
richfitz, line 24: as the number of things that wants to use this queue increases, overall time will go up. Any thoughts on how we should budget runners going forward?

docker/bin/test_self
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -e
+echo "--- Buildkite package :package:"
+R CMD build /src
+VERSION=$(grep "^Version" /src/DESCRIPTION | sed 's/Version: //')
```
richfitz, line 5: Consider also getting PACKAGE out of the package the same way, so no use of hintr, then this is reusable when we need it later

docker/coverage
```diff
@@ -28,6 +28,7 @@ docker run --rm --network=$NAME_NETWORK \
        -e VALIDATE_JSON_SCHEMAS=true \
        -e USE_MOCK_MODEL=true \
        -e _R_CHECK_CRAN_INCOMING_=FALSE \
+       -e TESTTHAT_CPUS=4 \
```
richfitz, line 31: this should detect the number of cores. Can we add something to our parallel runner to expose this? or detect cpu count on the runners (e.g., lscpu)

richfitz (changes requested):

R/zzz.R
r-ash: I think the default in traduire is to use `"translation"` as the default namespace when not set, so this change doesn't actually make a difference

buildkite/pipeline.yml
r-ash: Thinking we probably need some monitoring on the buildkite agents. Some idea of load, memory usage and idle time would be quite handy as a start so we can see if we can reduce the number of standard agents or if we need more parallel agents. I expect that in reality we have fairly spiky load where we might be using all agents for a short amount of time (e.g. because we have some builds which run many things in parallel) which might be quite challenging to handle. We don't want to end up with a bunch of buildkite agents sitting on all the resources on the server and being idle 95% of the time. I think getting some metrics would be a good start

docker/bin/test_self
r-ash: Thanks, that's much better

docker/coverage
r-ash: Nice, see https://github.com/reside-ic/reside-hyperv-scripts/pull/43

R/endpoints.R
r-ash: This is just because with mockr you cannot mock functions from other packages. As discussed let's go with this for now and can move to using stub and avoid `with_mock` in the future.

docker/Dockerfile
```diff
@@ -29,7 +29,8 @@ COPY docker/bin /usr/local/bin/
 # can hit rate limits as we cannot use a PAT when building the
 # container.
 RUN install_remote \
-    mrc-ide/naomi.options
+    mrc-ide/naomi.options \
+    mrc-ide/mockr
```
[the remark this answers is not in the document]
r-ash: Will leave a comment on a PR which was closed last year

richfitz (approved):

PR #449, opened 2023-01-16
mrc-3919: Add option `mock_model_trigger_error` which will force the mock model to error if TRUE

r-ash:
This is so we can test error handling without having to trigger an error from the full model output.

This is deployed on dev atm.


R/model_options.R
```diff
@@ -100,9 +100,32 @@ do_endpoint_model_options <- function(shape, survey, programme, anc) {
...
+        label = "Trigger mock model error",
+        controls = list(
+          list(
+            name = "mock_model_trigger_error",
+            type = "select",
+            help_text = "Set TRUE to force the model fit to error",
+            required = TRUE,
+            options = list(list(id = "true", label = "Yes"),
+                           list(id = "false", label = "No")),
+            value = "false"
```
richfitz, line 116: I'm surprised that this uses `"false"` here for FALSE, but presume that's correct. What's the serialisation approach?

R/run_model.R
```diff
@@ -24,6 +24,11 @@ run_model <- function(data, options, path_results, language = NULL) {
       )
     )
 
+    if (!is.null(options$mock_model_trigger_error) &&
+        options$mock_model_trigger_error == "true") {
```
richfitz, line 28: if there's a chance that this option might be missing in prod you might use `identical(opptions$mock_model_trigger_error, "true")`. Any reason not to be using TRUE/FALSE here though rather than strings?

R/model_options.R
r-ash: Yeah so for this "select control" type the values all have to be a string see https://github.com/mrc-ide/hintr/blob/master/inst/schema/ModelRunOptions.schema.json#L11 . It would be nice if we could have boolean types but would have to be updated in dynamic form in front end and we just went with doing the conversion from string to boolean here instead

R/run_model.R
r-ash: Mention half above but the dynamic form only works with strings in SelectControl so this will be returned to as a string atm. Would save us some conversion pain if we updated that to be able to take booleans too but probably something for another ticket

richfitz: definitely another ticket!

PR #451, opened 2023-01-30
Allow resource URL to be an explicit null in SessionFile for backwards compatibility

r-ash:
This PR will
* Allow `resource_url` in SessionFile schema to be null for backwards compatibility

Needed for https://github.com/mrc-ide/hint/pull/808/files


LekanAnni (approved):
LGTM 👍 

I built it on dev, works as expected.

PR #454, opened 2023-02-22
Unskip previously flaky test

r-ash:
I'm guessing this is failing because of some issue with test ordering, not very clear what that might be though. This seems to be working reliably when running a garbage collect


richfitz (approved):
better than a skip I guess. Do we have a lot of finalize-based teardowns (rather than on.exit/withr::defer based deterministic ones)?

r-ash:
> better than a skip I guess. Do we have a lot of finalize-based teardowns (rather than on.exit/withr::defer based deterministic ones)?

No other test has explicit call to `gc` like this. I think nothing explicitly has a finalize teardown except for the `queue` object

PR #465, opened 2023-05-17
Add branch pin

r-ash:


r-ash:
DRC are unblocked by using firefox so we don't need this temporary hack anymore!

PR #466, opened 2023-08-14
Add new endpoints for returning calibration result and metadata separately

r-ash:
This PR will add 2 new endpoints
* `/calibrate/result/metadata/<id>` to return the metadata and warnings from the `/calibrate/result/<id>` endpoint
* `/calibrate/result/data/<id>` to return only the data from the `/calibrate/result/<id>` endpoint

This is the first step in heading towards smaller data payloads. We want to get hint working when metadata is fetched separately from the data itself. Note this PR doesn't remove existing endpoint, we can do that now or later.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hintr
 Title: R API for calling naomi district level HIV model
-Version: 1.1.14
+Version: 1.1.15
 Authors@R: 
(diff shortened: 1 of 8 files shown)


richfitz (approved):
LGTM if the responses work for Lekan

LekanAnni (approved):
LGTM 👍

PR #468, opened 2023-08-21
mrc-4508: Save plot data as a duckdb database

r-ash:
This PR will
* Save plot data output as a duckdb database (the bulk of the work for that is done in the Naomi PR https://github.com/mrc-ide/naomi/pull/405
* Add path to the metadata response to tell Kotlin backend where the file is available on disk. Might be some things we want to make a bit more robust here - this is going to be dependent on where the kotlin and hintr contains have the data mounted into the container. So is dependent on the docker setup, I think this is at the same path at the moment and maybe we can insist that it must be but feels a little fragile that we are relying on that.

Also do we want any additional metadata here? Kotlin will want to know how it can slice the data, which is information available in the filters but let me know if you need that be moved around or anything?

Couple of important things to note. In the duckdb database the plot data is stored in a table called `data` it is the only table in the database. In the database there are columns which are needed for building the metadata here (like the label columns) but are not used in plotting. If you filter them in the Kotlin side and just read the columns you need that should be fast (duck db stores data in columnar format). Currently for plotting the old endpoint returns columns `area_id`, `sex`, `age_group`, `calendar_quarter`, `indicator`, `mode`, `mean`, `lower` and `upper`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,7 @@ Imports:
     ids,
     jsonlite (>= 1.2.2),
     lgr,
-    naomi (>= 2.9.10),
+    naomi (>= 2.9.11),
     naomi.options (>= 1.1.0),
     porcelain (>= 0.1.8),
     qs,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hintr
 Title: R API for calling naomi district level HIV model
-Version: 1.1.15
+Version: 1.1.16
 Authors@R: 
     person(given = "Robert",
            family = "Ashton",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hintr 1.1.16
+
+* Add a new endpoint `/calibrate/result/path/<id>` which will return the path to the calibration plot data output which can be read by web backend
+
 # hintr 1.1.15
 
 * Add two new endpoints in preparation for web backend accessing the result data on disk directly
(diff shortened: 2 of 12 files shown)


EmmaLRussell (commented):
> Add path to the metadata response to tell Kotlin backend where the file is available on disk. 

I might be misremembering, but didn't we say we were going to take that information out of the metadata response (the front end doesn't care about it), and either let hint back end find the db on disk in a known location by calibration id, or return the location from a separate endpoint which just returns that?

> Kotlin will want to know how it can slice the data

Isn't the front end going to tell it that, based on the metadata e.g. indicator value. filter value? Is it ok for us to make assumptions e.g. that the filter names in the metadata will match the column names in the db?

r-ash:
> > Add path to the metadata response to tell Kotlin backend where the file is available on disk.
> 
> I might be misremembering, but didn't we say we were going to take that information out of the metadata response (the front end doesn't care about it), and either let hint back end find the db on disk in a known location by calibration id, or return the location from a separate endpoint which just returns that?

Agh yes we did, I'd forgotten about that! I'll fix that in this PR.

> > Kotlin will want to know how it can slice the data
> 
> Isn't the front end going to tell it that, based on the metadata e.g. indicator value. filter value? Is it ok for us to make assumptions e.g. that the filter names in the metadata will match the column names in the db?

Yep, that's what I mean really. It's ok to make those assumptions, the `column_id` will indeed match the columns in the database and the filter values are built from the data so they should be usable.

richfitz (approved):

EmmaLRussell (approved):

M-Kusumgar (approved):

PR #470, opened 2023-09-22
Naomi readonly

r-ash:


M-Kusumgar (approved):
Just tested it! Have tested the error multiple times, get it a decent number of times with hintr master branch, do not get it at all with this branch! Nicely done!

edit: whoops just realised this is the wrong place for the review

PR #475, opened 2023-10-18
mrc 4651: Add a dummy AGYW tool download

r-ash:
This PR will
* Add AGYW download type

This will just download a dummy xlsx file for now but puts the structure in place so we can get started on impl on web side.

This should be merged after corresponding Naomi PR has been approved and merged https://github.com/mrc-ide/naomi/pull/411 and we have removed the branch pins here.


richfitz (approved):
LGTM once the pins/naomi are ok

PR #476, opened 2023-10-27
mrc-4673: Add tableMetadata to calibrate result metadata

r-ash:
This PR will
* Add `tableMetadata` into the `/calibrate/result/metadata/<id>` and `/calibrate/result/<id>` endpoints which returns info related how the new output table should be displayed

Couple of things to note
* I added this to both endpoints because was keeping it up to date as I am still using it for some of the testing. Let me know if changing that causes a headache on your end. I'll make a ticket to rip that endpoint out entirely as I think it should no longer be being used?
* Compared to the spec, I've called the `filterSelections` `selected_filter_options` as this is what we do in the other endpoints which specify selected filters. Though worth noting the format here is slightly different, only returning IDs whereas the others return IDs and labels. I think I'll make a ticket to tidy that up in other places because no reason to return the label as that is duplicated info
* @EmmaLRussell slightly different than spec where we said we'd add a new endpoint, this is putting the data into an existing one. Any objections to that?

### DIFF
--- a/R/run_model.R
+++ b/R/run_model.R
@@ -83,8 +83,9 @@ select_data <- function(data) {
 
 process_result <- function(model_output) {
   output <- naomi::read_hintr_output(model_output$plot_data_path)
-  metadata <- build_output_metadata(output)
-  table_metadata <- build_output_table_metadata()
+  filters <- get_model_output_filters(output)
+  metadata <- build_output_metadata(output, filters)
+  table_metadata <- build_output_table_metadata(output, filters)
   warnings <- list()
   if (!is.null(model_output$warnings)) {
     warnings <- warnings_scalar(model_output$warnings)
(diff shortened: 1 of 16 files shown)


tests/testthat/test-05-calibrate.R
```diff
@@ -147,6 +148,12 @@ test_that("can calibrate a model result", {
                  "anc_clients", "anc_plhiv", "anc_already_art",
                  "anc_art_new", "anc_known_pos",
                  "anc_tested_pos", "anc_tested_neg"))
+
+  ## Table metadata is returned
+  expect_equal(names(result$tableMetadata), "presets")
+  expect_length(result$tableMetadata$presets, 2)
+  expect_equal(names(result$tableMetadata$presets[[1]]),
+               c("label", "column", "area_id"))
```
M-Kusumgar, line 156: i believe that is why the build is failing
```suggestion
               c("label", "column", "row"))
```

M-Kusumgar (changes requested):

tests/testthat/test-05-calibrate.R
r-ash: Thanks for that

M-Kusumgar (approved):
Looks good to me!

R/run_model.R
```diff
@@ -108,6 +110,27 @@ build_output_metadata <- function(output) {
   )
 }
 
+build_output_table_metadata <- function() {
+  list(
+    presets = list(
+      list(
+        label = scalar(t_("TABLE_SEX_BY_AREA")),
+        column = scalar("sex"),
+        row = scalar("area_id")
```
M-Kusumgar, line 119: actually have been using this and testing the waters and I just realised, this needs to be `area`. `area_id` is the `column_id` of the filter not the `id`!
```suggestion
        row = scalar("area")
```

M-Kusumgar (changes requested):

inst/schema/TableMetadata.schema.json
```diff
@@ -0,0 +1,33 @@
...
+            }
+          }
+        }
+      },
+      "additionalProperties": false,
+      "required": [ "label", "column", "row" ]
+    }
+  },
+  "properties" : {
+    "presets" : {
```
EmmaLRussell, line 26: We don't the option to also include any default filter values outside the presets? Those will just be whatever the defaults are for the choropleth?

inst/traduire/en-translation.json
```diff
@@ -87,5 +87,7 @@
     "QUARTER_4": "Quarter 4",
     "VALIDATION_MULTIPLE_LEVELS_DETAIL": "year {{year}} has area levels {{levels}}",
     "VALIDATION_MULTIPLE_LEVELS": "Data can only be for regions at a single area level per year. In uploaded data {{detail}}.",
-    "FAILED_ZIP_REHYDRATE_SUBMIT": "Cannot load from this zip file, archive missing required information. Please regenerate output zip and try again."
+    "FAILED_ZIP_REHYDRATE_SUBMIT": "Cannot load from this zip file, archive missing required information. Please regenerate output zip and try again.",
+    "TABLE_SEX_BY_AREA": "Sex by area",
```
EmmaLRussell, line 91: Need fr and pt for these too? We'll be fetching output metadata when language selection changes, and updating view from that as we already do for other output metadata I assume?

EmmaLRussell (commented):
Looks good as far as  I can tell, pending Mantra's requested change..

inst/schema/TableMetadata.schema.json
M-Kusumgar: the defaults in the frontend are actually just the first values in the filters array :laughing: so we should be alright

M-Kusumgar (approved):
LGTM! (pending emmas comment)

inst/traduire/en-translation.json
r-ash: Thanks! Yes I do

PR #477, opened 2023-10-31
Add qs to duckdb migration code and script

r-ash:
This PR will add some migration code to migrate from v2.9.10 to 2.9.11 and a script to run it. I've setup the code in a way that we could extend with more migrations in the future, though will need a little work to actually switch out the function which does the migration work, but should be decent start point. (Thinking we'll have to write one to migrate to duckdb 1.0.0 whenever that is released). The migration is quite chatty, and it will save out a qs file with a log of what happened during running so we should be able to see why things might be skipped.

This is a little hairy in places. The actual migration is fairly simple, read a file into memory and write it back out. But there is lots of book keeping around it. Happy to chat through if helpful.

My plan for running this would be
* Locate a project with a duckdb fit, and one with a qs fit
* Kill the front end
* Do a backup (with privateer or otherwise)
* Run migration
* Restart app
* Check both projects still load

### DIFF
--- a/R/migrations.R
+++ b/R/migrations.R
@@ -61,6 +61,33 @@ migrate_task <- function(task_id, queue, to_version, dry_run) {
 
   new_res <- migrate(res, to_version, dry_run)
   if (!dry_run) {
+    ## rrq stores results using an object store
+    ## So when an rrq completes a job successfully it generates an R object
+    ## This R object is either serialized and stored in redis or if it is
+    ## big it can be offloaded onto disk. The object store manages this
+    ## and also manages reference counting to avoid saving
+    ## duplicate values into redis or to disk. Note that this is different
+    ## to how naomi outputs are stored on disk.
+    ##
+    ## When a job finishes, the R object is hashed. If that hash already
+    ## exists in the object store then we add a new reference to it
+    ## with tag equal to the task id. This returns the hash which can be
+    ## used to locate this object later.
+    ## We then save the has into redis at the `task_result` key for this task
+    ##
+    ## So in Naomi example what happens is,
+    ## run_model or run_calibrate return a hintr_output object
+    ## this is a list which contains things like plot_data_path,
+    ## model_output_path, version, warnings
+    ## This is hashed and saved into the object store. Then that gives
+    ## a hash back which we save as the task result.
+    ##
+    ## So to change the result saved for a task during this migration we
+    ## need to
+    ## 1. Add the new result into the object store, getting the hash back
+    ## 2. Save the hash into redis as the `task_result`
+    ## The object store is an internal part of rrq and not exposed so
+    ## we need to be a big naughty here use ::: to access it.
     store <- rrq:::rrq_object_store(queue$queue$con,
                                     r6_private(queue$queue)$keys)
     hash <- store$set(new_res, task_id)
--- a/R/migrations.R
+++ b/R/migrations.R
@@ -1,28 +1,33 @@
-run_migration <- function(queue, output_dir, from_version, to_version,
-                          dry_run = TRUE) {
-  output_dir <- normalizePath(output_dir, mustWork = TRUE)
+run_migration <- function(queue, log_dir, to_version, dry_run = TRUE) {
+  log_dir <- normalizePath(log_dir, mustWork = TRUE)
   tasks <- queue$queue$task_list()
   status <- queue$queue$task_status(tasks)
   completed_tasks <- tasks[status == "COMPLETE"]
   migrations <- lapply(completed_tasks, migrate_task, queue,
-                       from_version, to_version, dry_run)
+                       to_version, dry_run)
   summary <- lapply(migrations, function(migration) {
     list(
       id = migration$id,
       action = migration$action
     )
   })
+
+  time_now <- iso_time_str()
   summary <- do.call(rbind, summary)
-  summary_path <- file.path(output_dir, "summary.csv")
+  summary_path <- file.path(log_dir, sprintf("summary_%s.csv", time_now))
   message(sprintf("Saving summary csv %s", summary_path))
-  output_path <- file.path(output_dir, "output.qs")
   utils::write.csv(summary, summary_path, row.names = FALSE)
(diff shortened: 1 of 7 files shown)


r-ash:
Think I might have to migrate duckdb here too. There are a few duckdb files on the server which have been created with an old version. Will add that tomorrow.

R/migrations.R
```diff
@@ -0,0 +1,93 @@
...
+  migrations <- lapply(completed_tasks, migrate_task, queue,
+                       from_version, to_version, dry_run)
+  summary <- lapply(migrations, function(migration) {
+    list(
+      id = migration$id,
+      action = migration$action
+    )
+  })
+  summary <- do.call(rbind, summary)
+  summary_path <- file.path(output_dir, "summary.csv")
```
EmmaLRussell, line 16: I take it it's ok to write the summary directly into the output dir? It's not going to get confused with anything else?
Is it worth timestamping the filename in case you do successive runs and want to keep previous summaries? Same for the qs file (which is the chattier log?)

R/migrations.R
```diff
@@ -0,0 +1,93 @@
...
+  message(sprintf("Saving summary csv %s", summary_path))
+  output_path <- file.path(output_dir, "output.qs")
+  utils::write.csv(summary, summary_path, row.names = FALSE)
+  message(sprintf("Saving output qs %s", output_path))
+  qs::qsave(migrations, output_path, preset = "fast")
+  migrations
```
EmmaLRussell, line 22: Does this actually need to be the return value?

R/migrations.R
```diff
@@ -0,0 +1,93 @@
...
+}
+
+migrate_task <- function(task_id, queue, from_version, to_version, dry_run) {
+  message(sprintf("Migrating %s", task_id))
+  res <- queue$queue$task_result(task_id)
+  if (!naomi:::is_hintr_output(res) ||
+      !all(c("plot_data_path", "model_output_path") %in% names(res))) {
```
EmmaLRussell, line 29: Would this happen if an output had partially failed, or if it was a really old task?

R/migrations.R
```diff
@@ -0,0 +1,93 @@
...
+    message(sprintf("Not migrating %s, invalid output format", task_id))
+    return(list(
+      id = task_id,
+      prev_res = res,
+      action = "No change - not migrateable"
+    ))
+  }
+  if (!is.null(res$version) &&
```
EmmaLRussell, line 37: Why would res$version be null? Do you also need to deal with version is null case as not migrateable?

R/migrations.R
```diff
@@ -0,0 +1,93 @@
...
+  out
+}
+
+migrate <- function(res, new_version, dry_run) {
+  plot_data <- naomi::read_hintr_output(res$plot_data_path)
+  new_plot_data_path <- tempfile("plot_data",
+                                 tmpdir = dirname(res$plot_data_path),
+                                 fileext = ".duckdb")
+  if (!dry_run) {
+    naomi:::hintr_save(plot_data, new_plot_data_path)
```
EmmaLRussell, line 83: So the updated data is just getting saved out in place, and the migrations output_dir is just a place  to keep the migrations log?

EmmaLRussell (commented):
Can't really comment on the nitty gritty of this, just some daft questions about the logic... 

Killing the front end before running the migration is just so that you don't get any new files appearing while the migration is progressing? 

I don't really see what function `from_version` serves here - it's just going to migrate anything whose version is less than `to_version` isn't it?

R/migrations.R
r-ash: If it's a really old task, at one point we had different names in the struct.

R/migrations.R
r-ash: It can be NULL for some very old model fits also. The structure of these outputs has gone through a few iterations. Yep I should deal with that case as not migrateable - will add in a condition for that.

R/migrations.R
r-ash: Yeah, so the calling script creates a new directory for writing out this log into. I'll tidy up this name as it is a bit confusing, will call it `log_dir` as that really is what it is. I think definitely a good idea to timestamp those files though - I'll do that too. The qs file is the chattier one yeah.

R/migrations.R
r-ash: Nope, will update this to return the paths to the log files now they are having the time appended

R/migrations.R
r-ash: Actually had a look at this. We can deal with it if it doesn't have a version number. So as long as there is a `plot_data_path` we can migrate it.

R/migrations.R
r-ash: Yep so this gets saved out to same dir as wherever the `res$plot_data_path` is. Just using `tempfile` here to create a random sequence of characters for us to save. 

So if `res$plot_data_path` is at `/results/my_plot_data.qs` the `tempfile` call will give us path `/results/plot_data432323.duckdb`.


Updated data will be saved out to a new file, not overwriting the old one here. The old one gets deleted below on line 84.

r-ash:
> Killing the front end before running the migration is just so that you don't get any new files appearing while the migration is progressing?

Yep that's right, to stop someone trying to read the file when we are deleting it


> I don't really see what function `from_version` serves here - it's just going to migrate anything whose version is less than `to_version` isn't it?

Yep you're right, had put this in just thinking about being generic. Thinking if we have another migration in the future which we want to run we can write another function like

```
migrate_v1.2.3_v2.0.0
```

and then run the correct function based on the `from_version` and the `to_version` passed in i.e. have some switching out of the actual function that does the migration work. That function here is just called `migrate` atm. So that way we could re-use the reporting and checking version stuff and only have to write another `migrate` equivalent function. 

It then gets put into the log too, but I suppose we could have all that stuff with just a `to_version`. I'll rip that out.

EmmaLRussell (approved):

EmmaLRussell:
Approved this but just noticed the integration test is failing! - also it's included the review comments @r-ash added while we were looking at this on my desktop...

PR #479, opened 2023-11-08
Add test and fix to ensure that uploaded data to naomi from prerun is in duckdb format

r-ash:
Thanks for spotting this potential bug @M-Kusumgar, turns out it could have caused us a problem.

I've put in a guard against that now

### DIFF
--- a/R/prerun.R
+++ b/R/prerun.R
@@ -36,6 +36,19 @@ hintr_submit_prerun <- function(inputs, model_output, calibrate_output,
   assert_names(inputs, required, optional)
   assert_files_exist(inputs)
 
+  ## Make sure data is in format required by the web backend
+  ext <- tools::file_ext(calibrate_output$plot_data_path)
+  ## Would be nice to do this in some way that doesn't have knowledge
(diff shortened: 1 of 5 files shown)


R/prerun.R
```diff
@@ -36,6 +36,19 @@ hintr_submit_prerun <- function(inputs, model_output, calibrate_output,
...
+  ## Make sure data is in format required by the web backend
+  ext <- tools::file_ext(calibrate_output$plot_data_path)
+  ## Would be nice to do this in some way that doesn't have knowledge
+  ## of what is the "latest" data type.. as this is coupling over
+  ## hint and naomi quite tightly but should protect against science
+  ## creating the model output without using the `naomi:::hintr_save` util
+  if (ext != "duckdb") {
+    plot_data <- naomi::read_hintr_output(calibrate_output$plot_data_path)
+    new_plot_data_path <- tempfile("plot_data", fileext = ".duckdb")
+    naomi:::hintr_save(plot_data, new_plot_data_path)
```
r-ash, line 48: I should probably export this

richfitz (approved):

PR #482, opened 2023-11-29
Add label and id to table metadata for column and row

r-ash:
Change `column` and `row` in the table metadata from a string to an object with `id` and `label` fields. `id` used for finding the values in the raw data and the `label` is what should be displayed

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hintr
 Title: R API for calling naomi district level HIV model
-Version: 1.1.20
+Version: 1.1.21
 Authors@R: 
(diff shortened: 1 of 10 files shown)


M-Kusumgar (changes requested):
I think we still want the table header to be area, you will need to modify the `filters` portion of the `tableMetadata` and change the label of the filter to be "Detail" rather than the row label

r-ash:
> I think we still want the table header to be area, you will need to modify the `filters` portion of the `tableMetadata` and change the label of the filter to be "Detail" rather than the row label

Oops thanks, fixed that!

PR #491, opened 2024-01-18
mrc 4959 Refactor metadata for filters component

M-Kusumgar:
Build isn't currently passing but the code should be fine, going to sort out some test failures today but does this approach make sense. Essentially we want of this type (from the calls I have had with you guys):
```typescript
type Warning = {
    locations: string[],
    text: string
};

type FilterId = string

type StateFilterId = string

type FilterType = {
    id: FilterId,
    column_id: string,
    options: FilterOption[]
}

type FilterRef = {
    filterId: FilterId,
    label: string,
    stateFilterId: string
}

type PlotSettingEffect = {
    setFilters?: FilterRef[],
    setMultiple?: StateFilterId[],
    setFilterValues?: Record<StateFilterId, string[]>
}

type PlotSettingOption = {
    id: string,
    label: string,
    effect: PlotSettingEffect
}

type PlotSetting = {
    id: string,
    label: string,
    options: PlotSettingOption[]
}

type PlotSettingsControl = {
    defaultFilterTypes: FilterRef[],
    plotSettings: PlotSetting
}

type NewMetadata = {
    // this has all possible filter types i.e.
    // area_id, area_level, calendar_quarter, indicator, sex, age_group
    filterTypes: FilterType[],

    // this has extra info about indicators such as scale, format, accuracy, but
    // this will have the same length as the indicator filter type options
    // we just dont need everything in here in the filterTypes
    indicators: Indicator[],

    plotSettingsControl: {
        choropleth: PlotSettingsControl,
        barchart: PlotSettingsControl,
        table: PlotSettingsControl,
        bubble: PlotSettingsControl
    },

    warnings: Warning[]
}
```


R/zzz.R
```diff
@@ -40,3 +40,7 @@ hintr_init_traduire <- function() {
 hintr_translator_unregister <- function() {
   traduire::translator_unregister()
 }
+
+t_ <- function(...) {
+  traduire::t_(..., package = "hintr")
+}
```
r-ash, line 46: I should have done this in naomi but maybe worth putting a note at the top here that this is needed because when running tests it sometimes picking up the package environment as `testthat` instead of `hintr` as it should.

tests/testthat/setup.R
```diff
@@ -0,0 +1,13 @@
+library(mockery)
```
r-ash, line 1: I think we do not need this? `with_mocked_bindings` provided by `testthat`.

r-ash: Can you move these helper functions into `helper-hintr.R`. Functions you want to source before running tests are in `helper-` by convention and then testthat will source them as part of its setup.

r-ash: Ah I see using this library to remove the namespacing when creating a mock. In general I always use namespaced calls in R packages as they affect the users environment which can cause problems. This is fine here as just sourced in tests, but perhaps worth moving into `tests/testthat.R` with the other library calls.

r-ash (changes requested):
Thanks, this looks good after a quick skim through but I will look properly on Monday.

Just a couple of surface level things. Could you also bump the version number in the `DESCRIPTION` file and add a news item in `NEWS.md`

R/zzz.R
M-Kusumgar: done

tests/testthat/setup.R
M-Kusumgar: okay makes sense! will move the using library bit to `test/testthat.R` and helper functions to `helper-hintr.R`!

inst/schema/CalibrateMetadataResponse.schema.json
```diff
@@ -2,20 +2,29 @@
...
-    "plottingMetadata": {
+    "filterTypes": {
+      "type": "array",
+      "items": { "$ref": "FilterTypes.schema.json" }
+    },
+    "indicators": {
+      "type": "array",
+      "items": { "$ref": "ChoroplethIndicatorMetadata.schema.json" }
+    },
+    "plotSettingsControl": {
```
EmmaLRussell, line 13:
```suggestion
    "plotSettingsControls": {
```

EmmaLRussell (commented):
Can't really comment much on the R code, but the example metadata in the description of this PR is really useful, particularly the comments! I wonder if you should include this somewhere (in the README?), ideally with additional comments covering what the difference is between plot settings, plot settings effect and plot setting controls, what the different effect fields mean, what stateFilterId means etc... These probably all seem really obvious to you, but didn't to me, and maybe won't to some approaching the metadata for first time or after a break! 

In `PlotSettingEffect`, am I right in remembering that `setFilters` are just the filters to include as available when the effect is selected, `setMultiple` are the subset of those filters which will allow multiple values, and `setFilterValues` are the default filter values which the effect will select when it's first loaded (and these default values won't be changeable by the user if the filter isn't also in `setFilters`?). I think I find the `set..` prefixes a bit confusing, but I guess it's to do with this being an 'effect' and what the effect does... so I think it will be fine with a little bit of docs.

What does `defaultFilterTypes` in PlotSettingsControl do? Is that the filters you get if the plot doesn't actually have any settings defined e.g. choropleth? 

Also, in your description example `plotSettings` is defined as a single `PlotSetting`, but that should be multiple (you can have multiple dimensions of settings on a single plot)?

M-Kusumgar:
documentation thoroughly
validation on backend for defaultFilterTypes to contain reference by other effects

r-ash:
Superseded by #493

PR #495, opened 2024-03-06
Use github actions for tests

r-ash:
This PR will move to github actions. This is quite noisy as lots of tidying to do
* Adds lots of `normalizePath` calls so that tests pass on windows where paths with `\` where causing problems when converting to JSON
* Removed some now redundant packages (including mockr now that testthat has `with_mocked_bindings`), and added some imports which were missing
* Calling scripts with `Rscript` explicitly as these were failing to run on windows

Note that this breaks the continuous deployment to staging, I think we can add this again at a later point using github deployments but not critical for now and cost of that not running is quite low.


.github/workflows/R-integration-test.yaml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: ${{ matrix.config.os }}
+
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {os: macos-latest,   r: 'release'}
+          - {os: windows-latest, r: 'release'}
```
richfitz, line 22: Is it worth dropping windows/mac support entirely to simplify the build?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,69 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - id: extract_branch
+        name: Extract branch name
+        shell: bash
+        run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+
+      - name: Set up QEMU
```
richfitz, line 20: What is qemu for here?

.github/workflows/test-coverage.yaml
```diff
@@ -0,0 +1,72 @@
...
+          use-public-rspm: true
+          extra-repositories: https://mrc-ide.r-universe.dev
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::covr
+          needs: coverage
+
+      - name: Start Redis
+        if: runner.os != 'Windows'
```
richfitz, line 37:
```suggestion
        if: runner.os != 'Windows'
```

can be deleted if you want, coverage here running only on ubuntu

R/calibrate.R
```diff
@@ -10,7 +10,7 @@ run_calibrate <- function(model_output, calibration_options, path_results,
     })
   }
 
-  path_results <- normalizePath(path_results, mustWork = TRUE)
+  path_results <- normalizePath(path_results, winslash = "/", mustWork = TRUE)
```
richfitz, line 13: I suggest making a utility

```
normalise_path <- function(path) {
  normalizePath(path, winslash = "/", mustWork = TRUE)
}
```

and using that everywhere. Just as much noise in the diff but easier to deal with in future

scripts/build_test_data
```diff
@@ -15,8 +15,9 @@ invisible(lapply(paths, function(path) {
   }
 }))
 
-out <- system2("./scripts/run_model",
+out <- system2("Rscript",
```
richfitz, line 18: This is fine, but `callr::rscript` can make this more pleasant (or even just `callr:r` and drop the docopt wrapper)

richfitz (commented):
This looks great - main suggestion would be to question what the value of the windows/mac builds are here.

M-Kusumgar (approved):
Just had a quick skim through this one, works for me, pending rich's suggesstions

r-ash:
Closing this as have added #500 which is a bit tidier than this. We might want this PR at a later point if we need to get the tests to run on windows or mac machines

PR #496, opened 2024-03-07
mrc-5119: Improve error message handling

r-ash:
To deal with the case where errors from tidyselect are coming through without a message.

See notes in the ticket for context here. We've had an issue report with no error message, this is because the code here was just pulling out the message. This PR uses `conditionMessage` to build the message and strips any ANSI escape characters from it, which should work better with rlang errors.

### DIFF
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -390,17 +390,22 @@ download_submit <- function(queue) {
   }
 }
 
+get_download_result <- function(queue, id, error_message) {
+  res <- queue$result(id)
+  if (is_error(res)) {
+    msg <- api_error_msg(res)
+    hintr_error(msg, "OUTPUT_GENERATION_FAILED")
+  } else if (is.null(res$path)) {
+    msg <- t_(error_message)
+    hintr_error(msg, "OUTPUT_GENERATION_FAILED")
+  }
+  res
+}
+
 download_result <- function(queue) {
   function(id) {
     tryCatch({
-      res <- queue$result(id)
-      if (is_error(res)) {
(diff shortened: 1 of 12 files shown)


R/endpoints.R
```diff
@@ -393,11 +394,11 @@ download_result <- function(queue) {
...
-      if (is_error(res) || is.null(res$path)) {
-        msg <- res$message
-        if (is.null(msg)) {
-          msg <- t_("FAILED_ADR_METADATA")
-        }
+      if (is_error(res)) {
+        msg <- api_error_msg(res)
+        hintr_error(msg, "OUTPUT_GENERATION_FAILED")
+      } else if (is.null(res$path)) {
+        msg <- t_("FAILED_DOWNLOAD")
```
r-ash, line 401: Added a more specific error message here too

tests/testthat/test-endpoints-upload.R
```diff
@@ -47,7 +47,7 @@ test_that("api can upload input files", {
   res2 <- api$request("POST", "/internal/upload/input/survey_data.csv",
                       body = charToRaw(file))
 
-  expect_equal(res, res2)
+  expect_equal(res$body, res2$body)
```
r-ash, line 50: These have started failing since we added request_id into the headers. Different value for the header between `res` and `res2`. What I'm not sure about is why this didn't fail in the PR where I added those headers. But hopefully this should resolve it.

richfitz (approved):
Looks good to me, though reviewing on my phone!

PR #500, opened 2024-03-26
Move CI to GitHub actions

r-ash:
I've carved this out of #495 

This is the content of #495 without any of the changes to get the tests working on windows and mac. Want to leave those in a branch if we do want them at some point but agree probably not much value in those at the moment.


richfitz (approved):
So this PR means that auto deploy to staging is gone now? That makes some sense given that you're moving to a different (as yet undecided) deployment pattern soon.

r-ash:
> So this PR means that auto deploy to staging is gone now? That makes some sense given that you're moving to a different (as yet undecided) deployment pattern soon.

Yeah that's right, for now we'll have to deploy manually

PR #510, opened 2024-04-25
New calibrate metadata

r-ash:


R/metadata.R
```diff
@@ -118,6 +123,66 @@ get_barchart_settings <- function() {
...
+      setFilters = c(calibrate_only_settings,
+                     lapply(filterIds, get_filter_from_id)),
+      setFilterValues = list(
+        indicator = c("prevalence"),
+        period = get_filter_option_ids(filter_types, "period")[2],
+        sex = get_filter_option_ids(filter_types, "sex")[1],
+        age = c("Y015_049"),
+        calibrate_type = get_filter_option_ids(filter_types, "type"),
+        spectrum_region = get_filter_option_ids(filter_types, "spectrum_region")
+      ),
```
M-Kusumgar, line 146: is this necessary? the frontend will automatically pick everything if a filter is multiselect (which seems like what you want for `calibrate_type` and `spectrum_region`) and if it is single select then it will pick the first one, what you might be doing in `age` and `sex`, i think `indicator` and `period` are genuinely needed although not sure about why the current plot on production does that

if you do this then might be worth just hard coding the period id and no need to pass `filter_types` in, the options might potentially change and then youll have a different default without realising

R/metadata.R
```diff
@@ -118,6 +123,66 @@ get_barchart_settings <- function() {
...
+      )
+    ),
+    ## x-axis and disaggregate plot settings are not visible as users cannot
+    ## change these in the calibrate plot
+    plotSettings = list(
+      list(
+        id = scalar("x_axis"),
+        label = scalar(t_("OUTPUT_BARCHART_X_AXIS")),
+        options = list(get_x_axis_or_disagg_by_option("spectrum_region")),
+        visible = scalar(FALSE)
```
M-Kusumgar, line 158: minor suggestion, perhaps a very low priority ticket but would be good to be consistent with this, you have a `setHidden` effect so would be nice if this boolean was `hidden` instead of `visible` but super lower priority so fine to ignore

M-Kusumgar (approved):
am happy with this! looks good, small suggestions but are optional, think this should work fine

R/metadata.R
r-ash: Thanks, yep let's be consistent here

R/metadata.R
r-ash: Oh great yes, I'll remove those unnecessary ones. I think we want to keep period working off the `filter_types` as these can change depending on the data. So we want the 2nd one chosen which should be the same as the time 2 option they select in the model options, and potentially different for different countries.

PR #514, opened 2024-05-08
Add metadata for table

r-ash:
This PR adds metadata for the table. The important deviation from existing data here is that we now have a "customPlotEffect", this is some metadata used only by the plot itself and not used to update the filters. Perhaps should call this "customPlotMetadata" or something. Let me know what you think. Then we use in the table to indicate what we use for rows and column groupings.


M-Kusumgar (approved):
awesome i think the `customPlotEffect` is a good addition!

PR #518, opened 2024-07-23
Add ability to reconnect to redis after a certain health check interval has passed

r-ash:


M-Kusumgar (approved):
LGTM! shame azure doesnt have an option that just makes this work but guess they gotta save resources

PR #520, opened 2024-07-25
add documentation about metadata

M-Kusumgar:
Let me know what your thoughts are on this! Hopefully it is clear.


metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
...
+│   │   │   │   ├── id
+│   │   │   │   ├── label
+│   │   │   │   ├── effect # here
+│   │   │   │   │   ├── setFilters
+│   │   │   │   │   ├── setMultiple
+│   │   │   │   │   ├── setFilterValues
+│   │   │   │   │   ├── setHidden
+```
+
+For this section we will need a more detailed view of the effects structure
```
david-mears-2, line 214:
```suggestion
For this section we will need a more detailed view of the effects structure:
```

metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
...
+│   ├── filterId # 1.1
+│   ├── label # 1.2
+│   ├── stateFilterId # 1.3
+├── setMultiple[] # 2
+├── setFilterValues{} # 3
+├── setHidden[] # 4
+```
+
+Explanation of each effect and how the frontend aggregates this:
+1. `setFilters` is how the effect specifies what `filters` the user sees from the `filterTypes` The frontend aggregates this by replacing it. So if two `settings` set filters, the last `setFilters` in the for loop would be used when going through all the `settings`. Keys:
```
david-mears-2, line 227:
```suggestion
1. `setFilters` is how the effect specifies what `filters` the user sees from the `filterTypes`. The frontend aggregates this by replacing it. So if two `settings` set filters, the last `setFilters` in the for loop would be used when going through all the `settings`. Keys:
```

Adds full stop

metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
...
+    1. `label` is the title of the `filter` displayed to the user in the UI
+    1. `stateFilterId` is the id the state uses to identify this filter. This is distinct to the `filterId` because in the bubble plot we have two indicators filters that are the same filter type (so have the same `options` and the same `column_id`) however they are distinct because one controls the color of the bubbles and one controls the size of the bubbles so while both their `filterId`s are `indicator`, the `stateFilterId`s for both are different, `colorIndicator` and `sizeIndicator`. For most cases this can be set as the same thing as the `filterId`. All other effects will now identify filters by `stateFilterId` as these uniquely correspond to what the user sees
+1. `setMultiple` is an effect that sets a `filter` to a multi-select dropdown (they are all single select by default). It is an array of `stateFilterId`s (`1.3` in the visualisation) This is aggregated via concatenation in the frontend so if two settings have `setMultiple` properties `["state_id_1", "state_id_2"]` and `["state_id_3"]` respectively then all three filters with those state ids will be set to multi-selects
```
david-mears-2, line 231: What visualisation? Edit: Oh I see, the schema above, I was looking for an image.

david-mears-2: Slightly clearer to say "converts a `filter`" rather than "sets a `filter`" because I was trying to interpret the sentence like "sets" meant "applies".

metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
...
+│   ├── label # 1.2
+│   ├── stateFilterId # 1.3
+├── setMultiple[] # 2
+├── setFilterValues{} # 3
+├── setHidden[] # 4
+```
+
+Explanation of each effect and how the frontend aggregates this:
+1. `setFilters` is how the effect specifies what `filters` the user sees from the `filterTypes` The frontend aggregates this by replacing it. So if two `settings` set filters, the last `setFilters` in the for loop would be used when going through all the `settings`. Keys:
+    1. `filterId` is the `id` key in a `filterTypes`element (`a` in the visualisation)
```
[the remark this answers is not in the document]
david-mears-2: Oh I see, the schema above, I was looking for an image.

metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
...
+│   │   │   │   ├── effect
+│   │   │   │   │   ├── setFilters
+│   │   │   │   │   ├── setMultiple
+│   │   │   │   │   ├── setFilterValues
+│   │   │   │   │   ├── setHidden
+```
+
+The plot settings controls is an object with keys of the relevant plot name (e.g. `choropleth`) and value with the structure shown in the visualisation. Explanation of keys:
+
+1. `defaultEffect` is the base effect applied to all `settings` (remember these are the values of the `control` dropdown) - more on this later
```
EmmaLRussell, line 139: ...ok, so the default effect is like the baseline and the other effects describe how they deviate from the baseline?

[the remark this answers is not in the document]
EmmaLRussell: yeah, I've got confused again about how controls, settings, options and effects all fit together here.
Rather than just showing this structure and describing the parts of it, I think it would be really useful to screenshot a relevant part of the UI and work through it, saying which parts of the metadata contributed to what parts of the UI, and how e.g. changing a preset activates new bits of the metadata.

EmmaLRussell (commented):
Sorry, only got halfway through this before becoming completely confused, and you've already explained this to me at least once!

I really like the thoroughness of the explanation but I feel like starting from the metadata structure and trying to explain the individual bits of it doesn't really help my understanding. I remember when you explained it to me before in terms of the different use cases and why each part of the metadata was required to support each of them, that was really useful. I feel like an explanation which starts from the parts of the UI - ideally as screenshots with arrows showing which bits map to settings, controls, effects etc. in the corresponding metadata example - would make this all seem less abstract than it does currently. Sorry that would be quite a lot of work, but I feel like it would be worth it! 

Also, just a few style and clarity suggestions for the first part of text..

metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
+# Metadata driven UI
+
+## Motivation
+
+Historically Hint had different components for their plots each with their own logic on how to handle different filter selections. For example:
```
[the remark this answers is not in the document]
r-ash: I agree with this, no need to to document previous impl. I do think the list of features that we wanted the metadata to support is useful i.e. switching between mutliselect and single select and presets.

metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
...
+```
+
+Explanation of each effect and how the frontend aggregates this:
+1. `setFilters` is how the effect specifies what `filters` the user sees from the `filterTypes` The frontend aggregates this by replacing it. So if two `settings` set filters, the last `setFilters` in the for loop would be used when going through all the `settings`. Keys:
+    1. `filterId` is the `id` key in a `filterTypes`element (`a` in the visualisation)
+    1. `label` is the title of the `filter` displayed to the user in the UI
+    1. `stateFilterId` is the id the state uses to identify this filter. This is distinct to the `filterId` because in the bubble plot we have two indicators filters that are the same filter type (so have the same `options` and the same `column_id`) however they are distinct because one controls the color of the bubbles and one controls the size of the bubbles so while both their `filterId`s are `indicator`, the `stateFilterId`s for both are different, `colorIndicator` and `sizeIndicator`. For most cases this can be set as the same thing as the `filterId`. All other effects will now identify filters by `stateFilterId` as these uniquely correspond to what the user sees
```
r-ash, line 230: I guess the take home from this is that `stateFilterId` should be used as the primary key for the filter? So all effects reference it via its `stateFilterId`. The `fitlerId` should only be used to actually filter the data itself and to identify the `filterType`?

r-ash (changes requested):
This is great, thanks for writing this up. Couple of small comments I left, and a couple of other things.

1. Could you make this into a vignette, we can then put it onto a docs site in a later PR
2. Could you expand the bit about default effect, I can't quite remember a couple of things after not looking at this for a while. These get run when the payload is received right to set the initial state? Do they also get run when we change any of the plot controls is that right? In that way if we have an effect in the default and in the plot control then they interact as explained in the "Effects" seciton? As in `setFilters` will override the default, but `setMultiple` is additive?

metadata-structure/metadata-structure.md
```diff
@@ -0,0 +1,340 @@
...
+│   │   ├── description
+├── indicators
+├── plotSettingsControl{}
+│   ├── defaultEffect
+│   │   ├── setFilters
+│   │   ├── ... (see effect below)
+│   ├── plotSettings[]
+│   │   ├── label
+│   │   ├── id
+│   │   ├── visible
```
r-ash, line 165: This is `hidden`

metadata-structure/metadata-structure.md
r-ash: Updated this, the schema already limits this to just 1 per control but can't do that across multiple controls. Updating wording a little

vignettes/metadata.Rmd
```diff
@@ -0,0 +1,460 @@
...
+1. Add metadata into existing endpoint or new endpoint
+1. If a new endpoint, add an action which fetched the metadata, runs `filtersAfterUseShapeRegions` and `commitPlotDefaultSelections`.
+1. Add the new plot type into `PlotName`
+1. Add handling for new plot type into `getMetadataFromPlotName`, it should just map to the location of the metadata
+1. Add handling for new plot type into `getPlotData`, if this is using a new data source, write a function to do the filtering for this data source, and run any specific logic for this data source type.
+1. Write the component pulling data from `plotData` state and metadata via `getMetadataFromPlotName`.
+
+### Where do I write custom logic?
+
+**Modify the metadata before it is ever used**, write it in the action after the data is fetched and before it saved to state. For exampe, setting options for a filter to those from the shape file.
```
M-Kusumgar, line 452:
```suggestion
**Modify the metadata before it is ever used**, write it in the action after the data is fetched and before it saved to state. For example, setting options for a filter to those from the shape file.
```

PR #523, opened 2024-09-06
Remove old endpoints, add tests

r-ash:
This PR will
* Remove old unused endpoints
* Write tests for new endpoints

Note I left the time series endpoints as they are for now. When you tidy that up perhaps we should remove those? Though not too much harm leaving them in


R/endpoints.R
```diff
@@ -549,15 +549,7 @@ get_input_choropleth_data_source_options <- function(types, default_area_level)
 model_options_validate <- function(input) {
   input <- jsonlite::fromJSON(input)
   tryCatch({
-    ## Update some labels to match what naomi requires
-    ## TODO: Some of this is shared between model running and here so we
-    ## should use use common code when we merge this back into hintr.
-    ## This endpoint currently isn't called see mrc-592.
```
r-ash, line 555: This is unrelated, but just tidied this up whilst I was here

tests/testthat/helper-expect-metadata.R
```diff
@@ -0,0 +1,246 @@
...
+                         filter_ids,
+                         paste(context, "setFilters"))
+    state_filter_ids <- vcapply(effects$setFilters, "[[", "stateFilterId")
+    state_filter_map <- c(state_filter_map,
+                          setNames(filter_ids, state_filter_ids))
+    expect_state_filter_id_unique(state_filter_map)
+  } else if (length(state_filter_map) == 0) {
+    ## This is kind of gross but logic is, if setFilters is not included
+    ## in an effect, then use all the available filters. So have to replicate
+    ## this here for validation
```
r-ash, line 178: @M-Kusumgar this is true right?

tests/testthat/helper-expect-metadata.R
```diff
@@ -0,0 +1,246 @@
...
+  }
+  if (!is.null(effects$setFilterValues)) {
+    expect_filter_exists(names(filter_types),
+                         state_filter_map[names(effects$setFilterValues)],
+                         paste(context, "setFilterValues"))
+    for (state_filter_id in names(effects$setFilterValues)) {
+      filter_id <- state_filter_map[[state_filter_id]]
+      if (isTRUE(filter_types[[filter_id]])) {
+        break
+      }
```
M-Kusumgar, line 194: sorry bit confused by this, is this a truthy check on whether the key exists in `filter_types` and why do we break if the key does exist?

tests/testthat/helper-expect-metadata.R
```diff
@@ -0,0 +1,246 @@
...
+      all_option_ids <- get_filter_option_ids(filter_types[[filter_id]]$options)
+      expect_option_exists(all_option_ids,
+                           effects$setFilterValues[[state_filter_id]],
+                           paste(context, "setFilterValues"),
+                           effect$id)
```
M-Kusumgar, line 199: where is `effect$id` coming from?

M-Kusumgar (commented):
Mostly looks good to me! i had a couple of questions just to make sure im understanding this correctly!

tests/testthat/helper-expect-metadata.R
r-ash: agh this is some old junk from previous loop which I missed. Should be `state_filter_id` now, thanks.

tests/testthat/helper-expect-metadata.R
r-ash: o damn more silly mistakes here. This is supposed to be if `use_shape_regions` is true for this filter then skip the check of the options

tests/testthat/helper-expect-metadata.R
r-ash: Thanks! Should have looked at that code, thanks for the link. I've updated the test using proper information from here. This has simplified this function a bit too!

tests/testthat/helper-expect-metadata.R
```diff
@@ -0,0 +1,258 @@
...
+  if (length(set_filter_values) > 0) {
+    expect_filter_exists(names(filter_types),
+                         state_filter_map[names(set_filter_values)],
+                         paste(context, "setFilterValues"))
+    for (state_filter_id in names(set_filter_values)) {
+      filter_id <- state_filter_map[[state_filter_id]]
+      if (isTRUE(filter_types[[filter_id]]$use_shape_regions)) {
+        ## If we're referring to a filter which uses regions from
+        ## shape file, we have no options in metadata. So skip this check.
+        browser()
```
M-Kusumgar, line 202: rogue debug?

tests/testthat/helper-expect-metadata.R
```diff
@@ -0,0 +1,258 @@
...
+validate_effects <- function(filter_types, effects, default_effects, context) {
+  ## Effects are additive, except for setFilters where effects on an
+  ## individual filter override the defaults
+  set_filters <- effects$setFilters
+  if (length(effects$setFilters) == 0) {
+    set_filters <- default_effects$setFilters
+  }
+  expect(length(set_filters) > 0,
+         paste(context, "setFilters must be set in plot setting or",
+               "default settings."))
```
M-Kusumgar, line 174: while this is currently true for all plots, this is not necessarily true, feel free to leave this in, but you can technically do more with the metadata.

For example, say there was a plot control `A` and a plot control `B` and a default effect `C`. You can have it so that only `A` contains a `setFilters` effect for all its settings and `B`, `C` do not contain a `setFilters`. Then when your test gets to plot control `B`, it will fail this test but every option from plot control `A` still specifies `setFilters` so everything in the UI would be fine

if you want to properly test this behaviour i would go one level up, pull our combined `setFilters` (combined from all combinations of all available plot controls + default effect) and test that none of those combined `setFilters` are empty. that would test it out properly but up to you whether you want to do this or not

M-Kusumgar (approved):
happy to approve after that debug is removed, had an optional comment about testing the behaviour of the `setFilters` effect!

tests/testthat/helper-expect-metadata.R
r-ash: Thanks, added a note about this (feeling a bit lazy!)

M-Kusumgar: no problem!

PR #527, opened 2024-09-18
[nm-42] Make it impossible to accidentally destroy redis data

r-ash:
This PR will 
* Change the queue cleanup code so that data will not be deleted from redis unless you explicitly opt in

Should hopefully avoid recreating the accidental delete of all of our records which happened the other day :see_no_evil:

### DIFF
--- a/R/queue.R
+++ b/R/queue.R
@@ -3,7 +3,8 @@ Queue <- R6::R6Class(
   cloneable = FALSE,
   public = list(
     root = NULL,
-    cleanup_on_exit = NULL,
+    stop_workers_on_exit = NULL,
+    delete_data_on_exit = NULL,
     controller = NULL,
     results_dir = NULL,
     inputs_dir = NULL,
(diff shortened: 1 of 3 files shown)


R/queue.R
```diff
@@ -12,12 +13,14 @@ Queue <- R6::R6Class(
     next_health_check = NULL,
 
     initialize = function(queue_id = NULL, workers = 2,
-                          cleanup_on_exit = workers > 0,
+                          stop_workers_on_exit = workers > 0,
```
M-Kusumgar, line 16: i think this is still a bit dangerous because the default is creating 2 workers and it doesnt discriminate between different workers so if theres some on prod and we accidentally spawn new ones, itll still bring down all the workers on exit, `worker_manager$id` has all the ids of the workers spawned by this queue so i think we should also only kill those workers

r-ash: Thanks, that's a good shout. I've made those changes too now. Will not start any workers by default, and it will only cleanup the workers the `queue` creates itself.

scripts/redis
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 set -e
-CONTAINER_NAME=hintr_redis
+CONTAINER_NAME=hintr-redis
```
r-ash, line 3: Changing name here so that we can run up the front end dependencies at the same time as we run this. If we want to work on both at the same time.

M-Kusumgar (approved):
Nice :) hopefully will be much safer

PR #528, opened 2024-09-23
[NM-45] Add some validation to project rehydration endpoint

r-ash:
I was running into issues developing the new "open output zip from ADR" feature. And we occasionally get this with rehydrating an output zip. We should only get this if we are loading an output zip from another deployment. But at the moment when this happens the front end just kind of falls over. So returning a better error message which should mean that a project never gets created if state is invalid.

This PR will
* Return an error from rehydrate result if any input file is not found on disk
* Return an error from rehydrate result if model ID or calibrate ID are not found in redis


M-Kusumgar (approved):
looks good to me!

PR #534, opened 2024-11-08
[NM-61] Update population endpoint to return data and metadata for population pyramid

r-ash:
This PR will
* Return data and metadata from the `validate/baseline-individual` for the population file which we can use for building the population pyramid plots


krisdoyon (approved):
Metadata is working and has been integrated with population plot, this should be merged alongside the hint nm-62 branch

PR #553, opened 2025-03-03
[NM-115] Add ability to configure workers and queues via json

r-ash:
At the moment the worker and queue set-up is done in the package. I want to make this easier to modify on the move to the cloud because we're going to have to be more careful with resources.

With our cloud setup we can run on "consumption" or "workload" profiles, consumption is for <8GB of RAM and will spin up quickly, pretty much taking as much time as it takes to pull the docker container. Workload profiles for anything more than this basically spin up a VM and use this. So the startup time for this is larger.

So what I want to do, is to be able to configure per-country and per-job (model fit, calibrate, download etc.) where these run. And I want to be able to easily modify this during a workshop should we need to. (side note I've added a model option to force the fit onto the biggest node as Rachel and Jeff can bypass this should they need to - https://github.com/mrc-ide/naomi.options/pull/28)

So this PR does a few things
1. Make the workers and queue setup configurable via JSON, the config in the package here should be the same as what we are currently doing
2. I was getting an intermittent error from fread, it has been a long standing bit of work on my todo list to drop the usage of this in favour of readr so I have done this here too. This is what the naomi package is using for reading so should mean we have more consistent behaviour with them, and means we can cut down the size of the docker image a bit as we can remove the data.table depedency
3. Also added a small unrelated helper script for plotting average requests times from har files, I was using this for a bit of profiling comparison between current prod and azure deployment

## Follow up work
1. Update the deploy tool to use the new syntax for the entrypoint functions
2. Add ability to pull the configuration from azure config, and fallback to the one baked into the package if not set on the cloud
3. Add a wake up endpoint
4. Remove data.table from docker image build

### DIFF
--- a/R/worker_config.R
+++ b/R/worker_config.R
@@ -95,3 +95,12 @@ get_queue_from_job_name <- function(job_name, iso3 = NULL,
     queue_config$default
   }
 }
+
+validate_worker_name <- function(name) {
+  if (!(name %in% names(cfg$workers$workers))) {
+    stop(paste0("Cannot start worker with config '",
+                name,
+                "' this is not in configuration."))
+  }
+  invisible(TRUE)
+}
--- a/R/worker_config.R
+++ b/R/worker_config.R
@@ -1,0 +1,97 @@
+read_worker_config <- function() {
+  ## Read app config if we have this setup with env vars
+  config <- jsonlite::read_json(system_file("worker_config.json"),
+                                simplifyVector = TRUE)
+
+  ## Reverse the order of the job config so that lookups later are easier
+  ## we want something like list(job_name = list(country_iso3 = queue_name)))
+  config$job_mapping <- parse_and_validate_worker_config(config)
+  config
+}
+
+register_workers <- function(controller) {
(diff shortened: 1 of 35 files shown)


R/worker_config.R
```diff
@@ -1,9 +1,14 @@
 read_worker_config <- function() {
-  ## Read app config if we have this setup with env vars
-  config <- jsonlite::read_json(system_file("worker_config.json"),
-                                simplifyVector = TRUE)
+  worker_config <- Sys.getenv("HINTR_WORKER_CONFIG")
+  if (!is.null(worker_config) && worker_config != "") {
```
M-Kusumgar, line 3: rich taught me a trick, i think you can do
```suggestion
  worker_config <- Sys.getenv("HINTR_WORKER_CONFIG", unset = NA)
  if (!is.na(worker_config)) {
```
something like this

M-Kusumgar (approved):
thanks for talking me through it! looks like a flexible approach for scaling, looks good, have a  tiny comment about some R stuff but no biggie

PR #558, opened 2025-04-30
[NM-122] Pass PJNZ through to input time series programme data

r-ash:
Same as #557 but without all the metadata changes

So I did #557 adding a new "data source" in the drop down for comparisons but it was too weird and too much special casing. So I have ripped this metadata out in favour of this simpler approach, showing the adjusted indicators in the expanded view


r-ash:
Actually, will stick with #557

